Issue #20, opened 2023-12-06
Check debian package structure


igorpecovnik:
debian.conf defines paths mapping, which can be different on Git then on filesystem. See if I set i right and adjust. When done, I'll enable automated import to nightly repository.

Tearran:
As i understand.
all that is needed is 
- `configng/bin/* = /usr/bin/`
- `configng/lib/* = /usr/lib/`

Share folder at this stage is handled by [documents](https://github.com/armbian/configng/blob/main/lib/armbian-configng/documents.sh)  to overcome some naming conventions.

igorpecovnik:
> As i understand.

Yes

PR #21, opened 2023-12-06
Adjust quick install guide since we are making a repository on the fly

igorpecovnik:
After verifying if we put it into wanted folders and if it works, automation builds upon changes.


Tearran (approved):

Tearran (commented):
https://github.com/armbian/configng/blob/main/lib/armbian-configng/documents.sh#L206

Will Overwrite changes here

igorpecovnik:
So we want to have README generated via Github actions?

Tearran:
Automating the documentation is part or the project, since brand name issues or and compatibility remain concern within the community. Seemed prudent to use brand as naming convention. $filename --dev uses lib/$filename/* 

Where to direct it was unclear, so an option or Placeholder was made.  #20 cleared so A test to Redirecting to share/doc/$filename/  [documents](https://github.com/armbian/configng/blob/main/lib/armbian-configng/documents.sh#L206)  works locally Can delete the whole folder and armbian-confign will make folder share/doc/armbian-config 

## Naming conventions
### For example
`ln -s armbian-configng armbian-configng-dev`  will try and use `lib/armbian-configng-dev/*`  if it exist. 
` cp -r  lib/armbian-configng lib/armbian-configng-dev`  
currently wil output branded documents to banded doc folder.

### Alternatively
`unsupported-config` would look for `lib/Unsupported-config`

Tearran:
@igorpecovnik 

> So we want to have README generated via Github actions?

Not really, Was originally intended to have separate readme per `Category`.  In a doc folder.  This was the compromised that was made when it was agreed to direct focus away from [building and understanding ](https://github.com/Tearran/configng/tree/testing-deb_packages) how a *.deb are built. 

The doc generators can be easily directed to the share/doc folder now.

igorpecovnik:
ok. merging this?

Tearran:
Yes. Ran fine on a codespace. 

Not sure, about the ci/cd. I only spent few days to get familiar with what a ci/cd pipeline is.

igorpecovnik:
> Not sure, about the ci/cd. I only spent few days to get familiar with what a ci/cd pipeline is.

CI complains about the SheelCheck inconsistency and that has to be taken into account. For future. I also need to add to skip non bash files.

Issue #12, opened 2023-12-06
Is it possible to have Freeze / Unfreeze in the menu?


igorpecovnik:
Try freezing unfreezing kernel in the previous armbian-config

Tearran:
Existing config features can be accessed through Legacy menu.

https://github.com/armbian/configng/assets/2831630/ab56d4f9-cdb3-4132-8011-41ae233c0d3d

Tearran:
Loads key=pairs form `/etc/armbian-release`

```
{
    source /etc/armbian-release
    packages=("linux-image-current-$LINUXFAMILY" "linux-u-boot-$BOARD-$BRANCH" "u-boot-tools")

    for pkg in "${packages[@]}"; do
        if apt-mark showhold | grep -q "^$pkg$"; then
            sudo apt-mark unhold "$pkg"
            echo "Unheld $pkg"
        else
            sudo apt-mark hold "$pkg"
            echo "Held $pkg"
        fi
    done
}
```

Tearran:
Reference code migrated from current armbian-config [Job](https://github.com/armbian/config/blob/master/debian-config-jobs#L947) to make
 [Helper Function  ](https://github.com/armbian/configng/blob/main/lib/armbian-configng/config.ng.functions.sh#L193)  
[Job Enable](https://github.com/armbian/configng/blob/main/lib/armbian-configng/config.ng.jobs.json#L8) and [Job Disable](https://github.com/armbian/configng/blob/main/lib/armbian-configng/config.ng.jobs.json#L8)

Tearran:
Reference code migrated from current armbian-config [Job](https://github.com/armbian/config/blob/master/debian-config-jobs#L947) to make
[Helper Function](https://github.com/armbian/configng/blob/main/lib/armbian-configng/config.ng.functions.sh#L193)
[Job Enable](https://github.com/armbian/configng/blob/main/lib/armbian-configng/config.ng.jobs.json#L8) and [Job Disable](https://github.com/armbian/configng/blob/main/lib/armbian-configng/config.ng.jobs.json#L8)

PR #25, opened 2023-12-09
Function to freeze and unfreeze board firmware

igorpecovnik:
This function is an example (its just copy / paste from old and does not function due to missing helper functions) what we need to implement as common functions:
- whenever we are executing apt function, we need to check if apt is not working in the back
- we need to have all board / armbian related variables like ${BOARD} ${BRANCH} ...

Here I had in mind that menu would once says "Freeze" and second time "Unfreeze". We can solve this in several ways ... also we should condition this entry even file is there.


igorpecovnik:
https://github.com/armbian/configng/issues/12

schwar3kat:
> Here I had in mind that menu would once says "Freeze" and second time "Unfreeze". We can solve this in several ways ... also we should condition this entry even file is there.

Yes... this is exactly the type of thing that a menu initialization function in the script could achieve.  The menu could say freeze if it's not frozen (freezed) and could say unfreeze if it is frozen (freezed).

I'm not sure if my suggestions are viable or the best way to do it, but let's see if @Tearran can come up with a way to make this happen.

Tearran:
When posed with questions about How to, the communication breakdown becomes evident.

How to Check System State:

If the answer involves querying a specific file in /etc/x, this responsibility falls under the Frontend. It relates to user interface and tasks associated with interacting with the system state file.

If the answer is that you need a function to query the system to determine its state, this responsibility shifts to the Backend. It involves server-side logic and data processing for assessing the system state.

How to Toggle the System State:

If the answer is that you need a function to read the state and send a command to set it, this task aligns with the Communication Layer. It acts as an intermediary, managing the exchange of data and instructions between the Frontend and Backend.
By making these distinctions, we ensure a clear division of responsibilities:

Frontend: Bash scripts handling user interactions, including querying specific files for system state checks.

Backend: Bash scripts managing server-side logic, including functions to query the system state.

Communication Layer: Custom bash scripts or functions facilitating communication between the Frontend and Backend. It handles tasks like toggling the system state by reading the state and sending appropriate commands.

Tearran:
Ideally hoping the direction this is going that this repository will remain as the frontend while the corresponding function folders be migrated to the current config repo to further separate the front and back end aspects.

Tearran:
The current requirement for files in "lib/armbain-config/system/*.sh" 
- The header information
~~~
   # @description A short description of the function. `foundation for menus and cli commands`
   #
   # @exitcode 0  If successful.   `placeholder data setup working, no practice example of use`
   #
   # @options A description if there are options.  `placeholder data setup working, no practice example of use`
~~~
- run via the cli command. 
```
armbian-configng network=NMTUI
```
alternatively
```
armbian-configng --run network NMTUI
```
Image Description

While use case is not practical for cli is a practical example to outline current "function group" files, expectations and some limitations.


How do we differentiate menu topics and "function group" topics. 

examples could be. 
- Why does a menu need administrative privileges as a default
   - Configng does not require  administrative privileges to run
    - all previous examples show it is it a default requirement
- This is a menu setting `administrative privileges` as default was assumed a quick fix.
    -  --dev flag was added placeholder, $file-name-dev is supposed to address dev
- Why does a menu need to know exit code? 
    - No while loops.
- Why does a menu need to know arguments. or options? 
    - potential for sub sub menu

Or

- Why should a function group know if the menu need to be shown or not
  - `armbian-configng network=NMTUI`  bad example, still may have been all that was needed
- Why should a function group need know anything other than what commands to run
   -  ...
- Why is an cli style of `network=NMTUI ` required  when -o and --option arg is common
    - to facilitated communication layer from configng to ...

Tearran:
>Here I had in mind that menu would once says "Freeze" and second time "Unfreeze". We can solve this in several ways ... 

I agree, menu toggles were not overlooked.

>This function is an example , what we need to implement as common functions:

configng is not its library. They have separate responsibilities.  
- configng takes an input and parses it. 
- configng uses the the parsing array to provide help messages.
- configng has no external requirements to run. 
-  configng does not require administrator privileges. 

The libraries 
- lib/system most probable most if not all require admin privileges.    -  show if ran with sudo  and hide if none sudo user
- lib/network may or may not need sudo requires network individually hide sudo

if the input is network=NMTUI it will parse directly to the function 
if there is no input it is parses the the tui

The approach, How would one single bash script handel freeze unfreeze defined as system freeze that freezes ? 
any common functions can be placed in a central location in the root of lib/armbian-config after one is established. 

```
# @header ....
freeze(){
...
} 
unfreeze(){
...
}

kernal::Toggle(){

# check var or run soured function
if freezed && unfreeeze ; exit 0
if unfreeze && freese ; exit 0

}
```

Now that the "Legacy" was included in the main menu migration of functions research was on way.

>whenever we are executing apt function, we need to check if apt is not working in the back
we need to have all board / armbian related variables like ${BOARD} ${BRANCH} ...
 
- image-release has the context variables to continue

The start of the bottleneck and my focus on a documents generated from the header provided. 

Where are these variables located. READ the DOC
- [User-Guide_Armbian-Config](https://docs.armbian.com/User-Guide_Armbian-Config/) 
- [Developer-Guide_User-Configurations](https://docs.armbian.com/Developer-Guide_User-Configurations/)

Some other relative variable 
- [sbin/armbian-config](https://github.com/armbian/configng/blob/26-runtime-condition/lib/armbian-configng/runtime.sh) 
- some profile variable may have some relative variables as well

Tearran:
There is no traditional definition for "freeze" in this context. 

Previous assumption was to explore `read-only root filesystem` system and `read-write root filesystem`.


Current assumption `Hold/Unhold, u-boot and kernel updates`
Apt seems to be part of all builds, Uusing apt terms "hold" "unhold".  

```bash 
{
    source /etc/armbian-release
    packages=("linux-image-current-$LINUXFAMILY" "linux-u-boot-$BOARD-$BRANCH" "u-boot-tools")

    for pkg in "${packages[@]}"; do
        if apt-mark showhold | grep -q "^$pkg$"; then
            sudo apt-mark unhold "$pkg"
            echo "Unheld $pkg"
        else
            sudo apt-mark hold "$pkg"
            echo "Held $pkg"
        fi
    done
}
```
TODO trigger event communication layer. Failed but informative attempts have been made to get the relative information and provide doc generation of said variables last attempt was https://github.com/armbian/build/pull/5894 

Closed previous PR. Levering for for other reasons become a distraction. Abended idea of exposing any helpful information 

Exploring /share/armbian-config/configng.config key=pair for system state runtime checks. 
```
ku_freeze=$(apt-mark showhold)

[[ -z $ku_freeze ]] && #TODO change a Variable that can be read at runtime and change by group function
#TODO determine if the build first run can be leveraged to provide a default system state configuration file. 
```

The group function be responsible for changes system setting and to set the current state a file was intended. The config read this file and produce a json for other purposes, primarily  the gui and tui.

Tearran:
@igorpecovnik 
@schwar3kat 


The Idea was to have armbian-configng parse the parameters to the function and run command line. 
The date used to parse the parameter are also use (leveraged) to generate a json for a front end. 
 

## front end 

https://github.com/Tearran/retro-configng 


image

# Example of main menu and submenu
### Limitation
 - Whiptail is default
   - Dialog will only ever be partly compatible  

- Some conditions are still unknown, for run time.
- nesting is limited
- json manualy edit -  https://github.com/Tearran/retro-configng/blob/main/array.json
  - configng json generation is not compatible yet
  - Json generation dependent on unknown runtime conditions.       

# Example block category

```json
{
    "menu": [
      {
            "id": "Network",
            "description": "Network Options",
            "sub": [
                {
                    "id": "N1",
                    "description": "Wi-fi connect to network",
                    "command": "nmtui connect",
                    "show": true,
                    "requirements": ["nmtui"]
                },
                {
                    "id": "N2",
                    "description": "network manager tui",
                    "command": "nmtui",
                    "show": true,
                    "requirements": ["nmtui"]
                },
                {
                    "id": "N3",
                    "description": "Wi-fi connect to network",
                    "command": "nmtui connect",
                    "show": false,
                    "requirements": ["nmtui"]
                }
            ]
        }
    ]
}
```

Tearran (approved):
general concept applied

PR #28, opened 2023-12-10
fix for shellcheck error

Tearran:
Fixed some armbian-interface shellcheck errors and warnings

### DIFF
--- a/bin/armbian-interface
+++ b/bin/armbian-interface
@@ -36,7 +36,7 @@
 #
 
 ## DIRECTORY variable to the absolute path of the script's directory
-directory="$(dirname "$(readlink -f "$0")")"
+# directory="$(dirname "$(readlink -f "$0")")"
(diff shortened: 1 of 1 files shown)


bin/armbian-interface
```diff
@@ -36,7 +36,7 @@
 #
 
 ## DIRECTORY variable to the absolute path of the script's directory
-directory="$(dirname "$(readlink -f "$0")")"
+# directory="$(dirname "$(readlink -f "$0")")"
```
igorpecovnik, line 39: This is not needed anymore?

Tearran: Sellcheck says it is not needed, Since armbian-interface is an indepented app I would agree. 

at one time it was `filename=$directory...` and was forgotten after the chenge. 

Succefuly tested with the `echo "hi" | arbian-interfacce -o`

igorpecovnik (approved):

PR #36, opened 2024-05-19
Fix some typo's

schwar3kat:
Fixed a few spelling mistakes in Home.md


Tearran (approved):
The changes will need to be readdressed again in the jobs.json This will be overridden should be ok for now.

PR #38, opened 2024-07-07
Mordiaky patch 1

mordiaky:
Correct spelling in function.sh

"I had the time to take care of this as well :D let me know if you need anything else"


Tearran (approved):
Looks, great. You are Super star.

PR #39, opened 2024-07-08
Fix workflow

mordiaky:
removed unused folder in debian.conf
removed usr/share/ on line one


Tearran (approved):
Yes that is the changes needed. Thanks.

PR #45, opened 2024-08-29
Add jq whiptail and to default package dependencies

igorpecovnik:
# Description

When coming from minimal image, those are not installed but are needed.

# Testing Procedure

It shows no error but doesn't work and it works after this is installed.

# Checklist

### DIFF
--- a/.github/workflows/debian.yml
+++ b/.github/workflows/debian.yml
@@ -21,7 +21,7 @@ jobs:
       homepage: "https://www.armbian.com"
       section: "default"
       priority: "optional"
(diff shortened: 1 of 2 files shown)


Tearran:
@igorpecovnik Missing whiptial should show an error message. Seems there was a redundant check present. That should have been removed [#L286](https://github.com/armbian/configng/blob/main/lib/armbian-configng/config.ng.functions.sh#L286)

Also since checking for jq was overlooked added error message to check for both

Tearran (approved):
Seems like a good idea

PR #48, opened 2024-08-31
Install to internal storage - calling armbian-install script

igorpecovnik:
# Description

As title says.

# Implementation Details

Simple command execution. armbian-install is a part of BSP. Must be present ... do we need to check for presence?

# Testing Procedure

- [x] It opens and returns to armbian-config as expected

# Checklist

### DIFF
--- a/lib/armbian-configng/config.ng.jobs.json
+++ b/lib/armbian-configng/config.ng.jobs.json
@@ -74,7 +74,7 @@
                     "doc_link": "https://github.com/armbian/config/wiki#System",
                     "src_reference": "",
                     "author": "https://github.com/igorpecovnik",
-                    "condition": ""
+                    "condition": "[[ -n $(ls /sbin/armbian-install) ]]"
                 }
 
             ]
--- a/lib/armbian-configng/config.ng.jobs.json
+++ b/lib/armbian-configng/config.ng.jobs.json
@@ -63,7 +63,20 @@
                     "src_reference": "https://github.com/armbian/config/blob/master/debian-config-jobs#L160",
                     "author": "https://github.com/Tearran",
                     "condition": "are_headers_installed"
+                },
+                {
+                    "id": "S06",
+                    "description": "Install to internal storage",
+                    "command": [
+                        "armbian-install"
+                    ],
+                    "status": "Pending Review",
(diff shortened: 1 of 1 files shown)


lib/armbian-configng/config.ng.jobs.json
```diff
@@ -63,7 +63,20 @@
...
+                    "id": "S06",
+                    "description": "Install to internal storage",
+                    "command": [
+                        "armbian-install"
+                    ],
+                    "status": "Pending Review",
+                    "doc_link": "https://github.com/armbian/config/wiki#System",
+                    "src_reference": "",
+                    "author": "https://github.com/igorpecovnik",
+                    "condition": ""
```
Tearran, line 77: >Simple command execution. armbian-install is a part of BSP. Must be present ... do we need to check for presence?

No installation candidate in apt to check for armbian-install 
We could use
```json
 "condition": "[[ -n $(ls /sbin/armbian-install) ]]" 
```

Tearran (commented):
added "condition": "[[ -n $(ls /sbin/armbian-install) ]]"

Shows menu if armbian-install is present

to test renamed armbian-install.back does not show in menu.

Tearran:
Alternatively if there is an easy cli way to install armbian-install  
we can check if nvme or a 2nd mmc is available before showing menu and prompt user to install armbian-install if not present. 
```bash
#!/bin/bash

check_storage_devices() {
    local emmc_found=false
    local nvme_found=false

    # Check for eMMC devices (exclude boot partitions like mmcblk1boot0 and mmcblk1boot1)
    if lsblk -d -n -o NAME | grep -E '^mmcblk[1-9]$' > /dev/null; then
        emmc_found=true
    fi

    # Check for NVMe devices
    if lsblk -d -n -o NAME,MODEL | grep -E -i 'nvme' > /dev/null; then
        nvme_found=true
    fi

     echo $nvme_found

}
```
can check if an available device to install is available before showing menu item
```json
"condition": "check_storage_devices"
```

igorpecovnik:
> Alternatively if there is an easy cli way to install armbian-install

armbian-install is part of board support package, installed always. 

> we can check if nvme or a 2nd mmc 

It is more complicated ... i wouldn't bother with conditioning here.

Tearran (approved):

PR #55, opened 2024-09-04
Refactoring network configuration

igorpecovnik:
# Description

As we recently switched one level higher to defining network configuration via Netplan, all existing routines become deprecated. This commit adds initial support for handling wired and wireless networking. 

Problems - doesn't list sub routines:
`sudo bin/armbian-configng --cli N01`

jq: error (at bin/../lib/armbian-configng/config.ng.jobs.json:584): Cannot iterate over null (null)

Question:

What is simplest way to:

- pick and pass parameters further - IP, password
- collect users input via GUI  (reading IP, password, ...)

# Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code
- [ ] I have ensured that my changes do not introduce new warnings or errors
- [ ] No new external dependencies are included
- [ ] Changes have been tested and verified
- [ ] I have included necessary metadata in the code, including associative arrays

Issue #53, opened 2024-09-04
Propose small design change


igorpecovnik:
When we are running in CLI mode it would be desirable to have a switch to execute command without confirmation prompt.

Example:
`sudo bin/armbian-configng --cli S11`

This switch to stable repository and we can skip the prompt in machine mode. One way it would be to move this from command to new field

"prompt": [ "This will open interface file to edit\nCTRL+S to save\nCTLR+X to exit\nwould you like to continue?"

which we don't execute in CLI mode or if additional switch is used.

Tearran:
What is the purpose of CLI mode?

I assumed it is for CI/CD, GitHub Actions, or script integration. If this assumption is correct, then the --api flag is appropriate.

The current --cli flag should then be called --tui since it just bypasses menu selections, and --api should be invoked with --cli.

Tearran:
>This switch to stable repository and we can skip the prompt in machine mode. One way it would be to move this from command to new field

good idea
 implemented prompt 

example:
```json
"prompt": "This will open /boot/armbianEnv.txt file to edit\nCTRL+S to save\nCTLR+X to exit\nwould you like to continue?",
```


>When we are running in CLI mode it would be desirable to have a switch to execute command without confirmation prompt.
>
>Example:
>sudo bin/armbian-configng --cli S11

The only issue I see here is `--cli` flag may be misleading, `--cli` is `TUI` or `simple mode` 


 --api is intended for "cli mode" or "advanced mode"  passing command line arguments in a direct manner to avoid the nesting involved with the UI

igorpecovnik:
> flag may be misleading

Yes ofc. What about: --cmd or nothing? just -S11 ?

PR #60, opened 2024-09-09
network configuration

Tearran:
Added: 

- `Configure network interfaces (NetworkManager)` Menu sub section.
- input box to pass user data 
- Has --api (advanced mode) aka "CLI Mode" option to pass arg to change hostname or --cli (simple mode) to use the input box

## Issue reference:  
https://github.com/armbian/configng/issues/53  

## Pull request reference:
https://github.com/armbian/configng/pull/55

## Implementation Details

Configure network interfaces (NetworkManager) 

The existing nmtui was expanded to include other network setting for the TUI

I used for tui "simple mode"
```
sudo "./bin/armbian-configng" --cli SN07
```
I used  for "cli mode"

```
sudo "./bin/armbian-configng" --api network_hostname  khadasvim3
```

## justification 
This could be a good way to ensure that the basic network configuration is covered 
NetworkManager is still compatible with Netplan and included with desktop and Server images.  
```
[[ -n $( grep -m1 'renderer: NetworkManager' /etc/netplan/*.yaml) ]]
```

### DIFF
--- a/lib/armbian-configng/config.ng.jobs.json
+++ b/lib/armbian-configng/config.ng.jobs.json
@@ -283,46 +283,217 @@
             "description": "Wireless, Ethernet, and Network settings",
             "sub": [
                 {
-                    "id": "N00",
-                    "description": "Install Bluetooth support",
-                    "command": [
-                        "see_current_apt ",
-                        "debconf-apt-progress -- apt-get -y install bluetooth bluez bluez-tools",
-                        "check_if_installed xserver-xorg && debconf-apt-progress -- apt-get -y --no-install-recommends install pulseaudio-module-bluetooth blueman"
-                    ],
-                    "status": "Active",
-                    "doc_link": "",
-                    "src_reference": "",
-                    "author": "",
-                    "condition": "! check_if_installed bluetooth && ! check_if_installed bluez && ! check_if_installed bluez-tools"
-                },
-                {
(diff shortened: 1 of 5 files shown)


lib/armbian-configng/config.ng.jobs.json
```diff
@@ -283,54 +285,225 @@
...
-                    "command": [
-						"get_user_continue \"Verify that your Bluetooth device is discoverable!\" process_input ; connect_bt_interface"
-
+                    "id": "MN02",
+                    "description": "Configure network interfaces (NetworkManager)",
+                    "condition": "[[ -n $( grep -m1 'renderer: NetworkManager' /etc/netplan/*.yaml) ]]",
+                    "sub": [
+						{
+                            "id": "SN05",
+                            "description": "Select a Network Connection",
```
igorpecovnik, line 347: After introducing of Netplan networking config by default, we can drop all direct use of Network Manager. IP, DNS, DHCP ... everything is defined via NetPlan, which passes it to either networkd or network manager. We need to detect it and pass a parameter `renderer` correctly. This is also backward compatible.

function:
etplan_wrapper "10-dhcp-all-interfaces" "ethernets" "**networkd|NetworkManager**" "${SELECTED_ADAPTER}" "$IP_ADDRESS""

PR #110, opened 2024-09-24
Docker improvements and enabling auto testing

igorpecovnik:
# Description

Rework Docker install and enable automated testing routine.

# Implementation Details

Rework Docker install to use wget only. We don't need to use Curl and wget. When using BASH 4+ (which we do), we can use internal function to conduct _lowercasing_. Also improve purging by deleting the key and sources.

# Testing Procedure

- [x] Auto testing enabled and successful

# Checklist


lib/armbian-configng/config.ng.jobs.json
schwar3kat, line 1: LGTM

schwar3kat (approved):

PR #150, opened 2024-10-06
Add independent CPUFREQ config manipulation options

igorpecovnik:
# Description

Introducing new config `/etc/default/armbian-cpufrequtils` that is independent from deprecated cpufrequtils. Functionality only manipulate config file, while service needs to be added at buildtime.

# Implementation Details

A service that will read /etc/default/armbian-cpufrequtils needs to be added to armbian OS. Functionality should be merged in "Disabled" state.

# Testing Procedure

- [x] Configuration is generated and manipulated correctly

# Checklist

### DIFF
--- a/lib/armbian-configng/config.ng.functions.sh
+++ b/lib/armbian-configng/config.ng.functions.sh
@@ -212,6 +212,16 @@ function set_runtime_variables() {
 	DIALOG_CANCEL=1
 	DIALOG_ESC=255
 
+	# Generate empty CPU freq file in case it does not exists
+	if [[ ! -f /etc/default/armbian-cpufrequtils ]]; then
+	cat <<- EOF > "/etc/default/armbian-cpufrequtils"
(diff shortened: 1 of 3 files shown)


EvilOlaf:
Isn't cpufrequtils considered deprecated?

igorpecovnik:
Yes. This should be reworked.

igorpecovnik:
Now, we "only" need a replacement service to read those files and set CPU.

sp2ong:
Hi,
Is there a ready tool to set min/max cpu frequency. Is it still in testing status? very useful option that was once available.

igorpecovnik:
cpufreq which we used before, is deprecated / eol. This here is an replacement attempt. I am sure there must be several options but it would be best to keep this as minimal as this POC. You are welcome to try to implement it. This might help https://docs.armbian.com/Contribute/Armbian-config/

PR #180, opened 2024-10-13
161-feature-request-store-and-show-port-number

Tearran:
Cockpit 
image

Plex and emby

image
 

Theoretically, it should now be able to change the ports, and they will dynamically reflect the ports being used.


igorpecovnik (approved):
Looks good.

PR #247, opened 2024-11-11
[Template] Refine array and module parsing. Practice example portainer

Tearran:
Change port to expect defaults 
 
 google reddit reveals. 8000 (optional) and 9000, as the exspect ports for "CE"
 
 added reset 
 why: as to not have to re download image
 
add start stop 
why: needed for managing container 

# how it was tested 

` ./bin/armbian-config --api module_portainer help`
```

Usage:  <command>
Commands: help install uninstall start stop reset status
Available commands:
        status  - Show the status of the Portainer service.
        remove  - Remove Portainer.
        start   - Start the Portainer container.
        stop    - Stop the Portainer container.
        reset   - Reset the default state: Portainer container.

```

`./bin/armbian-config --api module_portainer stop`
`./bin/armbian-config --api module_portainer start`
`./bin/armbian-config --api module_portainer reset`


igorpecovnik:
> Change port to expect defaults

I think we will sooner or later clash into a problem where ports will be the same with different apps, thus representing a problem when both apps will be installed. I think we need internal port number scheme and add a field to the JSON database:
USED_PORTS="80,2000,1222" ... to verify port overlapping at development stages. Also when port is changed, documentation in tools/include/markdown has to follow.

igorpecovnik:
> option to pass values

I don't think this is the biggest problem. We need to have all ports in the database and choose where we will put some service too. Still user could have something installed on certain port, so we need to check ports before installing. This can quickly become messy ;) Lets first try to find out what do we really need and then code.

igorpecovnik:
> I would suggest the same approach

OK, this will do.

Since we are altering database, perhaps adding maintainer field so we make. When X adds Pi-Hole for example, he is maintainer while Y (Pi-hole repo or its founder) is author. And maintainers should be primary, author optional field. 

> perhaps include literally "other"

This part I don't quite understand. What do you mean by other here?

igorpecovnik:
>     ["module_portainer,options"]="9000 9002" # 2nd option

Do we expect anything else here? Otherwise its better to have `module_portainer,ports`

> ["module_portainer,maintainer"]="@armbian" # replace author 

["module_portainer,author"]="" # if different then maintainer

Perhaps?

Tearran:
>Do we expect anything else here? Otherwise its better to have module_portainer,ports

In short maybe! There is an optional "edge mode" that uses port 8000

Long answer, this is not as much about this module as it is about all modules in general. module_portainer is the flagship and basis of the modules_name(){..} for software. 

The idea here was to have `positional parameter` with `potential arguments.` other words,  Dynamic and reusable array

Perhaps we need something more reusable and optional 
```
["module_portainer,options"]="port" # options
["module_portainer,args00"]="9000 9002" # args
```

| description | module_option | command |  option[0]  | args[0] |
|---|---|---|---|---|
| | required | required | optional | optional | 
| | $0 | $1 |  $2 |  $3 | $4 | $5 | $6 | $7 |
| installs default | module_portainer | install | 
| installs specific port  | module_portainer | install | port | 9000 |

```
["module_portainer,options"]="port edge unknown" # options
["module_portainer,args00"]="9000 9002" # args
["module_portainer,args01"]="8000 8080" # args
["module_portainer,args02"]="foo bar bliss" # args
```
| description | module_option | command |  option[0]  | args[0]  |  option[1]  | args[1]  |  option[2] | args[2]
|---|---|---|---|---|---|---|---|---|
| | required | required | optional | optional | optional | optional |
| | $0 | $1 |  $2 |  $3 | $4 | $5 | $6 | $7 |
| installs default | module_portainer | install | 
| installs specific port  | module_portainer | install | port | 9000 |
| installs edge port  | module_portainer | install | port | 9000 | edge | 8080 |
| installs edge port  | module_portainer | install | port | 9000 | edge | 8080 | unknown | bliss |

igorpecovnik:
> Perhaps we need something more reusable and optional

Key is that anyone that is not you and me also understand the concept right away. But keeping it simple is always hard ...

Ports. Application is using more then one port and as it is running inside container, we have  port mapping in place, not just ports in use. From our position, we need to check that those externals does not overlap across applications. For this reason I am looking out rather to have ports as standalone property and leave random args for exotic exceptions.

Tearran:
Added port to module_option and quick port change to next in list. 
Added some known ports for a baseline
quick check for conflicts
```
 ./bin/armbian-config --api module_list_ports
```

```
module_portainer:
9000
9001
9002
Conflict: Port 9002 is used by multiple modules
module_list_ports:
80
443
22
21
587
124
module_cockpit:
9090
9091
9002
Conflict: Port 9002 is used by multiple modules
module_webmin:
10000
```
famailer with cockpit so used it for basic port switch
after installing cockpit 
```
./bin/armbian-config --api module_cockpit port
```
port    - Cycle fallback ports 
```
Cockpit socket configuration updated and restarted on port 9091
```

Tearran:
@igorpecovnik The simplest manner config.runtime.sh

igorpecovnik:
OK, port we need to have in help / docs too.

Issue #242, opened 2024-11-10
[Task]: Interesting and valuable titles / features to add to configurator


igorpecovnik:
### Task description

- [ ] Kodi
- [ ] Chromium (port from MESA extension)
- [ ] GoogleChrome (we host at our repo)
- [ ] Firefox (we host non snap version at our repo)
- [ ] LAMP stack with CERT bot
- [ ] Samba Server (enable, disable, config)
- [ ] NFS Client (enable, disable, config)
- [ ] Nextcloud
- [ ] XRDP
- [ ] proFTPD
- [ ] Home Assistant (taking from here https://github.com/armbian/configng/pull/235/files)
- [ ] Syncthing
- [ ] WireGuard
- [ ] Sonarr
- [ ] Radarr
- [ ] Grafana
- [ ] OctoPrint
- [ ] VSCodium (hosted at our repo)
- [ ] VSCode (hosted at our repo)

dimitry-ishenko:
If nobody is working on Samba, I can give it try.

EvilOlaf:
Jellyfin might be interesting, especially for rk3588 based devices since they actually support this platform.

EvilOlaf:
My observations when setting up atuin server and testing:  

After initial installation the config file isn't written directly but is created after `register`.  
When registering with `ATUIN_SYNC_ADDRESS=https://services.armbian.de/atuin` this value is not written back in the freshly generated config and needed adjustment.  
Therefore when implementing this via configng the config file with adjustments needs to be written the installation (line 32: `sync_address = "https://services.armbian.de/atuin"`)

EvilOlaf:
What about wirelogd?
https://github.com/nikaro/wirelogd

igorpecovnik:
> What about wirelogd?

Yes. First step - adding to our repo https://github.com/armbian/os/blob/main/external/gh.conf

> don't have permission.

Try now
 
> *rr stack is WIP

This is ready, tested, all works, won't be adding anything more there for this section.

> I suggest to remove this and add as module for configng:

Agree. There is also this as an option https://docs.linuxserver.io/images/docker-wireguard/?h=wire

igorpecovnik:
Yes, I saw. I think it can be valuable, lets add it.

igorpecovnik:
> If nobody is working on Samba, I can give it try.

Did you have time to look into this? Samba is pretty high on the usage interest level.

dimitry-ishenko:
[...]

Issue #266, opened 2024-11-22
[Task]: Collections of small future improvements


igorpecovnik:
### Task description

Let's assemble a list first, coding starts when this note is removed.

- [ ] Show actual IP address in the dialog box after installing applications. Currently it shows: `https://<your.IP>:8123`
- [ ] 
- [ ] 
- [ ]

Tearran:
>note: The distinction from `interactive mode` `non-interactive mode`. Was somewhat lost with the divergent keys of the json.  modules are built around module_options `database`

The database need attention. the keys are outdated such as "feature" was replaced with "id" in the json object and other keys were added. 

Ideally Before adding to the `Database` it need to be updated to reflect the clearer meaning `json` `keys`. 

### basic/interactive mode
`--cmd` previously `--cli` , user testing and a shortcut to the menu title 
`armbian-config --cmd  Benchy`
`armbian-config --cmd MAN005`

### advanced/non-interactive mode
`--api` option for `unit testing` modules, outputting` db info` into `json, md, config/` etc. 
`armbian-config --api module_webmin help` 
`armbian-config --api generate_json_options`

Tearran:
To dynamically  change the menu sub sub menu titles  `update_sub_submenu_data` in `config.runtime.sh` 

Adding a Data point should not break things. but will need to refactor the doc and generate_json_options. 
```
["module_name,port"]="<port number>"
```
Updating the menu to show the port 

```
update_sub_submenu_data "ID" "SUB_ID" "SUB_SUB_ID" "$LOCALIPADD:${module_options["module_name,port"]}"
```
SImple verify 
example  cockpit and webmin
```
# Sub sub menu updates
cockpit_port=$(lsof -iTCP:9090 -sTCP:LISTEN | grep systemd )
if [[ -n "$cockpit_port" ]]; then
	update_sub_submenu_data "Software" "Management" "MAN004" "https://$LOCALIPADD:9090"
fi
#
webmin_port=$(lsof -iTCP:10000 -sTCP:LISTEN | grep webmin )
if [[ -n "$webmin_port" ]]; then
	update_sub_submenu_data "Software" "Management" "MAN005" "https://$LOCALIPADD:10000"
fi
```
containers would be something like `| grep docker-pr `

Tearran:
### Existing keys 

- [x] id 
- [x] sub  
- [x] description
- [x] about
- [x] command
- [x] status
- [x] condition
- [x] example

# Suggested keys
- [ ] port
[...]

PR #272, opened 2024-11-27
Add Samba package

dimitry-ishenko:
This is a draft PR to make sure I am on the right track.

@igorpecovnik in #242 you have _enable_, _disable_ and _config_ options. What did you have in mind exactly?

Also, couple of issues I've noticed so far:

1. I've had issues with `debconf-apt-progress` not being able to uninstall `samba` with the following error:
```
Removing samba (2:4.19.5+dfsg-4ubuntu9) ...
Setting up samba-common (2:4.19.5+dfsg-4ubuntu9) ...
Can't locate object method "new" via package "Text::Iconv" (perhaps you forgot to load "Text::Iconv"?) at /usr/share/perl5/Debconf/Encoding.pm line 65, <GEN2> line 2.
dpkg: error processing package samba-common (--configure):
 installed samba-common package post-installation script subprocess returned error exit status 255
dpkg: dependency problems prevent configuration of samba-common-bin:
 samba-common-bin depends on samba-common (= 2:4.19.5+dfsg-4ubuntu9); however:
  Package samba-common is not configured yet.

dpkg: error processing package samba-common-bin (--configure):
 dependency problems - leaving unconfigured
Processing triggers for man-db (2.12.0-4build2) ...
Errors were encountered while processing:
 samba-common
 samba-common-bin
E: Sub-process /usr/bin/dpkg returned an error code (1)
```
Installing `libtext-iconv-perl` handled the error...

2. I've noticed that noticed that most (all?) packages use `apt-get purge` to uninstall themselves. The problem with this is that it does not uninstall dependencies. I've used `apt-get autopurge` instead, which add `--autoremove` option to remove all dangling dependent packages.

### DIFF
--- a/tools/json/config.software.json
+++ b/tools/json/config.software.json
@@ -354,6 +354,28 @@
                             "status": "Stable",
                             "author": "@armbian",
                             "condition": "check_if_installed avahi-daemon"
+                        },
+                        {
+                            "id": "NET009",
+                            "description": "Install Samba",
+                            "command": [
+                                "get_user_continue \"This operation will install Samba service.\nDo you wish to continue?\" process_input",
+                                "debconf-apt-progress -- apt-get -y install samba"
+                            ],
+                            "status": "Stable",
+                            "author": "@dimitry-ishenko",
+                            "condition": "! check_if_installed samba"
+                        },
+                        {
+                            "id": "NET010",
+                            "description": "Uninstall Samba",
+                            "command": [
+                                "get_user_continue \"This operation will purge Samba service.\nDo you wish to continue?\" process_input",
+                                "debconf-apt-progress -- apt-get -y autopurge samba"
+                            ],
+                            "status": "Stable",
+                            "author": "@dimitry-ishenko",
+                            "condition": "check_if_installed samba"
                         }
                     ]
                 },


Tearran:
Good start. The commands should be in a module.  
https://github.com/armbian/configng/blob/main/tools/template/module_software_template.sh

function module_samba() {

}

igorpecovnik:
> What did you have in mind exactly?

I didn't use correct wording. II meant install / uninstall and configuration. We need to insert some default config. People usually export some common data folder as read only accessible without password. [App download section](https://github.com/armbian/configng/pull/270) is using /armbian/ as a common download folder. IMO that should be available by default upon enabling (installing) samba. Anything else?

About issues.

1. I haven't used Samba in years, but I assume this part should be easy to resolve. We could be looking at different install commands at Debian and Ubuntu based rootfs. I hope not, but this kind of exceptions can be handled easily when a dedicated module for this feature is made. As proposed by @Tearran too. Even we are looking at simple apt commands, we should make a module.
2. Agree. Here I wasn't sure what is best practice, so I opened another ticket https://github.com/armbian/configng/issues/266 where task is to unify those procedures. I have added this there.

PR #276, opened 2024-11-29
Adding a warning to welcome screen

Tearran:
# Description

Added check_os_status compares /etc/armbian-distribution-status to os uses displays ok/exit window. 
used the color module `set_color 1`  
>NOTE: manually changed /etc/armbian-distribution-status to `bookworm=` `eos`, `tested`, and `csc` for test. 
Result
image

if = `supported` operates as normal


tools/modules/functions/check_os_status.sh
```diff
@@ -0,0 +1,42 @@
...
+		help)
+			echo "Usage: check_os_status"
+			echo "This function checks if the current OS is supported based on /etc/armbian-distribution-status."
+			echo "It retrieves the current OS distribution and checks if it is listed as supported in the specified file."
+			;;
+		*)
+			FILE="/etc/armbian-distribution-status"
+
+			# Detect the current OS distribution codename
+			CURRENT_OS=$(lsb_release -cs 2>/dev/null || grep "VERSION=" /etc/os-release | grep -oP '(?<=\().*(?=\))')
```
igorpecovnik, line 21: Isn't this already known at this stage?
https://github.com/armbian/configng/blob/main/tools/modules/functions/set_runtime_variables.sh#L52-L62

tools/modules/functions/check_os_status.sh
```diff
@@ -0,0 +1,42 @@
...
+			echo "Error: Unable to detect the current OS distribution."
+			exit 1
+			fi
+
+			# Check if the OS is listed as supported in the file
+			if grep -qE "^${CURRENT_OS}=.*supported" "$FILE"; then
+			echo "The current OS ($CURRENT_OS) is supported."
+
+			else
+			BACKTITLE="Error: The current OS ($CURRENT_OS) is not supported or not listed"
```
igorpecovnik, line 34: This warning should be telling that this tool has been verified on $SUPPORTED distributions, Armbian + bookworm, jammy, noble at this stage ... and if not any of those ... tool might still work well, but you are on your own Something like that.

tools/modules/functions/check_os_status.sh
Tearran: $DISTROID
```
No LSB modules are available.
bookworm
```
$CURRENT_OS
```
bookworm
```

igorpecovnik: yeah, fix that first one. no need to execute this twice

Tearran:
image

to test commented out line for bookworm. `/etc/armbian-distribution-status`

igorpecovnik:
Yeah, something like that.

igorpecovnik:
Adjusted text a bit - check if it still sounds o.k.

Also there is some grep issue - on exit:
```
grep: : No such file or directory
User canceled. exiting
```

Tearran:
Removed file "/etc/armbian-distribution-status" to replicate message.
```
grep: : No such file or directory
User canceled. exiting
```  
This is happening on an `unsupported os`  or a `modified os` ( such as I did to test it ) That is aso unsupported as I understand. IMO Fixing this error seems a little counter productive.

igorpecovnik:
> grep: : No such file or directory

I would suppress this error - we anyway show that conditions are not met with a big prompt.

Tearran:
@igorpecovnik redirected `grep: : No such file or directory`  to null


If `User canceled. exiting` is issue, Would like to addressed separately. note: its is part of `get_user_continue` and `process_input.`

PR #287, opened 2024-11-30
[Fix] module generate_json_object

Tearran:
# Description
Update and refine generate_json_option  
Planned early development. quick look as far back as PR reference:  #35, 
As part of the goal of an independent drop in module.

# Plan
- [x] Update generate json options.
- [x] parse/output a updated jq object. `./bin/armbian-config" --api generate_json_options` 
- [ ] Fix missing/malformed `Header` data points in the module_option arrays.
- [x] output a valid menu `generate_top_menu` object 
- [ ] Fix/update missing/malformed keys in `./Document.md` parsing

## Optional
- [ ] add new `Header` data points. such as `["..,port(s)"]="..."`
- [ ] add ["..,group"]="<parent id>" "Containers" or  "Media"

# Documentation Summary
To be able to catch up on the changes, a single data point should be used. The `header` is the main data point.  This provides a direct way to test and validate module_option(s) `Header`


Tearran:
`"./tools/config-assemble.sh" -p`
 ```
 "./bin/armbian-config" --api interface_json_data
```

some header info is malformed or missing data.  the selected for example 

image

PR #299, opened 2024-12-03
manage_dtoverlays: use overlays instead of fdt_overlays in armbianEnv…

amazingfate:
We don't have fdt_overlays in armbianEnv.txt, instead we are always using overlays.


igorpecovnik:
First problem nailed down, thanks.

I have a feeling we have more then this problem with overlays, right?

igorpecovnik (approved):

amazingfate:
`overlay_prefix` is not implemented now, so for rockchip-rk3588 vendor branch, panthor overlay is not able to get choosed.

PR #308, opened 2024-12-05
Kernel switching module: complete refactoring and hardening

igorpecovnik:
# Description

- current kernel switching management had several design flaws and we could end up without a kernel. 
- switching between stable and beta was pulled into this common modules.

# Implementation Details

BASH module.

# Documentation Summary

```
Usage: module_armbianfirmware <command> <switches>
Commands:  select install show hold unhold repository help
Available commands:
	select    	- [stable]|[rolling] TUI to select Armbian firmware.
	install   	- Install Armbian firmware [$BRANCH] [$VERSION].
	show      	- Show Armbian firmware packages.
	hold      	- [status] Mark Armbian firmware packages as held back. When parameter status is used, it returns false or true.
	unhold    	- Unset Armbian firmware packages set as held back.
	repository	- [stable]|[rolling] selects repository and performs update.

```

# Testing Procedure

- [ ] Tested all possible scenarios several times.

# Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code


igorpecovnik:
@Tearran @dimitry-ishenko @EvilOlaf 

Those changes above works well now, but there is a problem that menu got quite slow (Odroid M1 takes several seconds, my desktop is fast, so it gets <1s) due to many checks. Which are IMO needed to perform as designed. Or I am blind and I don't see how to do it better ;)

Alternatively adding sub menu here `System -> Kernel management` and put there:

- headers install
- hold upgrades
- rolling repository
- alternative kernels
- device tree overlays

Any thoughts?

Tearran:
> Alternatively adding sub menu here System -> Kernel management and put there:

for the front end and grouping. Adding header ["...,group"]="some_subgroup" could be easily used

PR https://github.com/armbian/configng/pull/287 Shows how we can query and nest separate groups. The example I used naming conventions module_ and gererate_ for groups others then fall in helper group.

```
"./bin/armbian-config" --api interface_json_data
```
image

EvilOlaf:
Nothing obvious. Though I am not a bash wizard. Maybe add a warning "can take several seconds" :D

igorpecovnik:
> Though I am not a bash wizard. Maybe add a warning "can take several seconds" :D

It takes too much of time already on arm64 board and this would go into 15 seconds or more in some old H3 board.

Here is a UX question first. If I push those kernel management stuff into submenu, main system will remain smooth as butter.

> ow we can query and nest separate groups. 

will look into this, thanks.

Tearran:
```json
                {
                    "id": "Kernel",
                    "description": "Manage Armbian kernel upgrades",
                    "command": [
                        "see_menu module_armbian_firmware"
                    ],
                    "status": "Stable",
                    "author": "@igorpecovnik",
                    "condition": ""
                }
```

Tearran:
condition checking in the help message to produce a list available options. https://github.com/armbian/configng/blob/main/tools/modules/software/install_webmin.sh

Tearran:
As shown with https://github.com/armbian/configng/blob/hardening-kernel-switch/tools/modules/software/install_webmin.sh some condition checking done when a module is needed instead of checking conditions all at once. 

"armbian-config" `->` `System` -> `Kernel`
`armbian-config --cmd Kernel`
`armbian-config --api see_menu module_armbian_firmware`

All produce 
{CF3520FC-D8A6-4F30-9C2D-0ED542D1E66A}

`armbian-config --api module_armbian_firmware`

The second tabs break the array. 
{D92785B3-25C9-466A-A7C1-0C668156F021}

igorpecovnik:
> The second tabs break the array.

I wasn't sure how to document those switches. Also here direct translation to the menu might not work well, at least how it is designed now - switch has more additional options. Kernel headers have remove, install, status (which is in this kind of arrangement not needed).

Tearran:
if you remove the 2nd \t on each line.  That will fix the strings.

You can nest the Kernel in a sub_group but suggest that a [...,group"]="Kernel" and   [...,conditions]="",   be added to the module_options along with [...,arch]="" and port even if not used.

igorpecovnik:
I think I will get back to this in the future, but for now I would merge good working but slow functioning. As kernel management was moved into submenu, this gets less annoying. And probably UX wise, this is also better.

PR #312, opened 2024-12-11
nfsd fixes

dimitry-ishenko:
# Description

Added a few small fixes for the nfsd module while reviewing it for my Samba work. 

# Implementation Details

- [ ] Key changes introduced by this PR

1. Added a small wrapper script to abstract away dealing with systemd services. I work inside a container (using systemd-nspawn) so all commands attempting to start/stop/restart a service fail.

2. Fixed `status` command which didn't work inside the container due to failing `systemctl is-active` command. Plus, checking if the service is active is not the same as checking if it was installed.

3. Fixed `uninstall` command, which should be `remove` instead.

- [ ] Justification for the changes

See above.

- [ ] Confirmation that no new external dependencies or modules have been introduced

Confirmed.

# Documentation Summary

- [ ] **Metadata Included:**  
 
None.

- [ ] **Document Generated:**  

None.

# Testing Procedure

Installed and uninstalled nfsd inside the container a few times.

# Checklist

### DIFF
--- a/tools/json/config.system.json
+++ b/tools/json/config.system.json
@@ -152,7 +152,7 @@
                             "id": "NFS03",
                             "description": "Remove network filesystem (NFS) daemon",
                             "command": [
-                                "module_nfsd uninstall"
+                                "module_nfsd remove"
                             ],
                             "status": "Stable",
                             "author": "@igorpecovnik",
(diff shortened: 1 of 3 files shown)


igorpecovnik:
Thank you! This will be useful elsewhere too.

I already went further here, so I cherry-picked your commits into my branch.

PR #320, opened 2024-12-12
Add wrapper for package manipulation

dimitry-ishenko:
What do you guys think about abstracting away package installation and removal? 

This will help add consistency and reduce boilerplate. Usage is as follows:

```
package install vim
package remove mc
package is-installed nmap
etc.
```

This will also replace `check_if_installed` and `apt_install_wrapper` commands.


Tearran:
@dimitry-ishenko This could be use to consolidate and refine or replace some of the following. Your idea is a good. 

https://github.com/armbian/configng/blob/3740e7dfe37f9866dfd804d23cb40899c3f83807/tools/modules/functions/check_if_installed.sh#L12

https://github.com/armbian/configng/blob/3740e7dfe37f9866dfd804d23cb40899c3f83807/tools/modules/functions/config_interface.sh#L428

https://github.com/armbian/configng/blob/3740e7dfe37f9866dfd804d23cb40899c3f83807/tools/modules/system/apt_install_wrapper.sh#L13

Reference for helper function naming [standards](https://github.com/armbian/configng/wiki/Standards), 
I suggest
```
["module_packages,example"]="help install remove update upgrade check"
``` 
or 
```
["set_packages,example"]="help install remove update upgrade check"
```

>note 
 IMO, Seems the inconsistency more a matter of a defining a procedure to follow then automating and document it. This is difficult while the remains database is incomplete, Part of the database is json and part associative array . The module_option array is missing data points that exist in the json. That need to be added to all the module_options.

dimitry-ishenko:
Thanks @Tearran. Are you suggesting to rename the file (ie, `package.sh` -> `module_packages.sh`) or also change the name of the function? I chose the name `package` because it makes the code natural and easy to read.

By the standard it would have to be something like:

```
get_package vim
rem_package mc
see_package_installed nmap
```

IMHO these are a bit less natural, but if that's what we gotta do, then that's what we gotta do.

Tearran:
Just the function name the prefix is used to separate groups. So if this is a module use module_ if it's a helper to be used with a module see_ set_ even etc.. 

This I would consider both so your call but only one prefix name. 

We can even refine and define new prefix to clarify things. We do have "generate_"  , Perhaps, "helper_" "manage_".

dimitry-ishenko:
> Just the function name the prefix is used to separate groups. So if this is a module use module_ if it's a helper to be used with a module see_ set_ even etc..
> 
> This I would consider both so your call but only one prefix name.
> 
> We can even refine and define new prefix to clarify things. We do have "generate_" , Perhaps, "helper_" "manage_".

TBH none of the prefixes really fit here... How about adding a `package_` or `pkg_` prefix? Then we would have:

```
package_install foo
package_remove bar
package_is_installed baz
```
or
```
pkg_install foo
pkg_remove bar
pkg_is_installed baz
```

Tearran:
package_ should work well. although IMO pkg_ fits the 3 letter prefix better, but that is unimportant ;) 

would also suggest simply "check" or "status" for the is_installed.  
```
package_install foo
package_remove bar
package_check baz
```

Quick fix : "Disabled" or ""
```
	["package_<job name>,status"]=""
```
If you don't mind helping refine module_option. 
The module_option status, may be misleading as it is. Last check "Disabled" is the only string that is used. 


Reference:
https://github.com/armbian/configng/blob/0a257fda5e14b5a5719478568a426cf4b27fa405/tools/config-assemble.sh#L158
&&
https://github.com/armbian/configng/blob/0a257fda5e14b5a5719478568a426cf4b27fa405/tools/modules/functions/config_interface.sh#L107

 I think we can leverage this to define what the function is.
```
	["package_<job name>,Helper"]=""
```
  For modules it would be. System, Network, Localization, Software.  and any other category that may be needed at later date.
#321 is about finishing the database then replace some slow bash parsing with JQ (json query).

dimitry-ishenko:
> package_ should work well. although IMO pkg_ fits the 3 letter prefix better, but that is unimportant ;)

Great idea, I will change it to `pkg_`.

dimitry-ishenko:
> If you don't mind helping refine module_option.
> The module_option status, may be misleading as it is. Last check "Disabled" is the only string that is used.

I am still getting a feel for what does what, where and how 😃  but I can take a look.

dimitry-ishenko:
@Tearran should I keep the name of the file as `package.sh`?

Tearran:
no standard set as for the logic inconsequential, so snakecase "any_name_will_do.sh" 

Software does set a president as it uses the prefix "install_"   
If we want to follow this naming consistency  IMO "helper_" would be appropriate.

Tearran:
@dimitry-ishenko in short code wise currently it does not matter can leave it if you like. But :) this is a good place to clarify, and refine exiting president for consistency.  

tested all works with  `--api` flag. 
Seems like Non interactive UX so perhaps should be "interface_" for file prefix. 
 
```
./bin/armbian-config --api package_install whiptail
```
{EB4B9D6B-308F-40A0-9631-9FB3C94A0365}

President 

https://github.com/armbian/configng/blob/main/tools/modules/functions/interface_menu.sh

The `[[ -t 0 ]]` Does Not work with the --api option

dimitry-ishenko:
This should be ready to go. Works in non-interactive mode as well.

dimitry-ishenko:
This should be ready to go. Works in non-interactive mode as well.

I've opted for `pkg_installed` instead of `pkg_check` or `pkg_status` because "check" or "status" can mean many different things. For example, this reads like almost natural English:
```
if pkg_installed zfsutils-linux; then
    ...
fi
```

Tearran:
Tested working 
`/bin/armbian-config --api pkg_install whiptail`
`/bin/armbian-config --api pkg_ uninstall whiptail`

`./bin/armbian-config --api pkg_update`
`E: The update command takes no arguments`

did not test upgrade

>Note There are 33 file with the command `check_if_installed`  These will need to be migrated as well. 
 being module_option related Most of it can be discussed and worked on here #321   

Quick fix:  stage for removal. 
- `["check_if_installed,status"]="Deprecated"` 
- `["apt_install_wrapper,status"]="Deprecated"` 

Disable being the only trigger string.  The helper will remain available for use and be document for removal.

Tearran (approved):
_update and _upgrade need some refining. manual test are promising and don't seem to break anything. Great direction

Tearran:
@igorpecovnik IMO nice improvements ready to go, any option? 
@dimitry-ishenko  amazing. even deprecated that like "pssst it was but a thing", lol NIce.

igorpecovnik:
Sorry for tuning in late. I was busy in other segments those days ... Great work! I would remove function tools/modules/functions/check_if_installed.sh right away so we don't drag it along anymore if its not used.

igorpecovnik (dismissed):
I assume it was replaced everywhere - in this case, free to merge.

dimitry-ishenko:
@igorpecovnik I've removed both deprecated functions, but it looks like I need another review before I can merge.

Tearran:
Something in the Unit tests  broke.

igorpecovnik:
> Something in the Unit tests broke.

It is usually unrelated. I am working to stabilize that it won't break for 3rd party reasons that much. You can probably re-run them too - if you select unit tests here in PR.

image

dimitry-ishenko:
Looks like all tests are good now. 🎉 If someone wants to merge, feel free (I can't do it without @igorpecovnik's approval).

igorpecovnik (approved):

dimitry-ishenko:
Thanks @Tearran

PR #321, opened 2024-12-12
[Fix missing] module_option mising option

Tearran:
# Description

The idea from the start was to have one data entry point, the "Header". The active data array can be viewed with `git repo` or `apt installed` using the command: `armbian-config --api generate_json_object`

This PR is the first step towards producing a JSON object pipeline that will eventually replace the JSON file. By leveraging the already used `module_option` array and `jq` for querying, we aim to ensure consistency and eliminate the need to manually edit or source a separate file.

Many JSON keys are missing in the `module_option` array and cannot be parsed or queried. Leveraging `jq` can:

# Changes Made

- Added missing keys to the `module_option` array for better parsing and querying.
- Updated the `generate_json_data.sh` script to properly handle the new keys and generate a valid JSON object.
- Modified the `config.runtime.sh` module_options to remove unhandled escape sequences and quotes.

# Goals and Impact

- Eliminate the JSON editing and automate the UX.
- Speed up the nesting that bash is unequipped for.
- Handle real-time changes better than bash alone.
- Ultimately ease automated and redundant tasks.
- Improved loading times and nesting issue handling.
- Better automation and less manual intervention required in JSON handling.
- Enhanced user experience through `jq` optimization.

# Testing

- Verified that the updated `./bin/armbian-config --api set_json_options` produces the correct JSON output.

# Notes

- `armbian-config --api generate_json_options` is out of date and needs updating.
- The `module_options` array was lacking needed keys, and this PR addresses multiple issues such as loading times and nesting issues that bash is not meant to handle.


Tearran:
Converts the module_option array into a raw flat json. for jq parsing.  
list all modules and helper functions that have headers. Included a "short id" . 
```
./bin/armbian-config --api set_json_data
```

 JQ  ("JSON Query")  generates json list of modules
```
./bin/armbian-config --api generate_json_data
```

Tearran:
{F9F6B5EF-1F1C-4314-91E7-6771F5FBCDCC}

Tearran:
Using existing database, we can output a useable software sections
 
{D1F66659-7B61-4DEB-BF4B-E97C3CDDFA4E}

{EE3C0E10-55E6-4BA9-84CE-9204E029C344}

Tearran:
{494DD53E-9F44-40DF-9964-D1AF48A17E3C}

used top menu setting for size

{6957FEAE-F31C-488C-9375-F6C83D999E99}

tools/template/module_software_template.sh
```diff
@@ -1,12 +1,18 @@
 
 
+
 declare -A module_options
 module_options+=(
 	["module_template,author"]="@Tearran"
+	["module_template,ref_link"]="@armbian"
```
igorpecovnik, line 7: what should go here? Maintainer or maintainer field should also be added?

Tearran: ref_link was hard to clarifi, I was thinking "reference link" would be for 3rd party homepage. if none armbian docs.

igorpecovnik: Hmm, we have doc_link below too. I think this is then doubled. Not sure we need links to armbian documentation as they are auto-included.

Tearran:
Closing

PR #339, opened 2024-12-25
Adjust software $json_data to match recent changes

Tearran:
# Description

The idea from the start was to have one data entry point, the "Header". The active data array can be viewed with `git repo` or `apt installed` using the command: `armbian-config --api generate_json_object`

This PR is the first step towards producing a JSON object pipeline that will eventually replace the JSON file. By leveraging the already used `module_option` array and `jq` for querying, we aim to ensure consistency and eliminate the need to manually edit or source a separate file.

Many JSON keys are missing in the `module_option` array and cannot be parsed or queried. Leveraging `jq` can:

# Changes Made

- Added missing keys to the `module_option` array for better parsing and querying.
- Updated the `generate_json_data.sh` script to properly handle the new keys and generate a valid JSON object.
- Modified the `config.runtime.sh` module_options to remove unhandled escape sequences and quotes.

# Goals and Impact

- Eliminate the JSON editing and automate the UX.
- Speed up the nesting that bash is unequipped for.
- Handle real-time changes better than bash alone.
- Ultimately ease automated and redundant tasks.
- Improved loading times and nesting issue handling.
- Better automation and less manual intervention required in JSON handling.
- Enhanced user experience through `jq` optimization.

# Testing

- Verified that the updated `./bin/armbian-config --api set_json_options` produces the correct JSON output.

# Notes

- `armbian-config --api generate_json_options` is out of date and needs updating.
- The `module_options` array was lacking needed keys, and this PR addresses multiple issues such as loading times and nesting issues that bash is not meant to handle.


Tearran:
Cleaning up outdated draft.

PR #345, opened 2024-12-27
Corrected torrent servers to torrent clients

zejjnt:
# Description

Really minor thing but Deluge, qBittorrent and Transmission are torrent _clients_, not torrent _servers_ :)

### DIFF
--- a/tools/json/config.software.json
+++ b/tools/json/config.software.json
@@ -428,8 +428,8 @@
                     "sub": [
                         {
                             "id": "DOW001",
-                            "description": "qBittorrent BitTorrent server ",
-                            "about": "This operation will install qBittorrent BitTorrent server",
+                            "description": "qBittorrent BitTorrent client ",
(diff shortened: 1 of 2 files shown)


Tearran:
IMO both are incorrect. It is a "peer to peer" network and acts as both server and client. So simply (P2P) or peer would be more accurate.

EvilOlaf:
> IMO both are incorrect. It is a "peer to peer" network and acts as both server and client. So simply (P2P) or peer would be more accurate.

Technically spoken yes. But in common linguistic usage these programs are known as _Clients_, therefore I agree with the suggested changes

EvilOlaf (approved):

Tearran (approved):
Client is more accurate,  2 of the 3 BitTorrent apps, [transmission](https://github.com/transmission/transmission) and [deluge](https://deluge-torrent.org/) call themselves clients.

Issue #349, opened 2025-01-01
[Bug]: Unexpected command executed when config overlays


atmgnd:
### What happened?

When managing overlays, If there is spaces in /boot/armbianEnv.txt, then the line will been executed as command, which is wrong

### How to reproduce?

Put some spaces /boot/armbianEnv.txt, for example

```
extraargs=apparmor=0 usbcore.autosuspend=-1
verlays=sun50i-a64-i2c0 sun50i-a64-i2c1
```

then run armbian-config, goto manage overlays. When entering and exiting overlay submenu, you can see that there will be errors in printing

![Untitled](https://github.com/user-attachments/assets/89eb3393-c1b3-41eb-9686-6f6df0b122de)

I think it caused by https://github.com/armbian/configng/blob/66d8921d8d12507b24efe6cfc1d241e4e2cca857/tools/modules/system/manage_dtoverlays.sh#L22 , in function manage_dtoverlays

```
[[ -f "${overlayconf}" ]] && source "${overlayconf}"
```

After remove it, no error printed anymore

### On which host OS are you running the build script and observing this problem?

Ubuntu 24.04 Noble

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Tearran:
@atmgnd Thank you for the report. 

Replicated error, a few times with different overlays, final was.
<img width="765" alt="{9D1C0A4F-CCD9-4498-8645-6D1C2DA01FDB}" src="https://github.com/user-attachments/assets/0de3d39a-6f36-4d34-8506-d80ad36f13de" />
```bash
/boot/armbianEnv.txt: line 7: meson-uartC: command not found
```

>[[ -f "${overlayconf}" ]] && source "${overlayconf}"
After remove it, no error printed anymore

Confirmed: No errors and still makes the correct edits to `/boot/armbianEnv.txt`.

PR #355, opened 2025-01-04
[Bugfix]:  Unexpected command executed when config overlays #349

Tearran:
remove problem line that sources `/boot/armbianEnv.txt` no obiouse adverce effets. 

issue: https://github.com/armbian/configng/issues/349


Tearran:
added a message for system that no not have /boot/armbianEnv.txt or the expected device tree overlay directory. 

{686B5A87-1AA0-4F6A-8CA7-CD0CD521272C}

igorpecovnik (approved):

PR #363, opened 2025-01-08
"Updates" menu not shown

torte71:
# Description
The menuentry "Updates" (under "System") is not displayed, even though the actual distro (bookworm) is supported.
This is caused by comparing the current distroid to a wrong part of the entries in `/etc/armbian-distribution-status`.

# Implementation Details
The first patch corrects the wrong field taken from `/etc/armbian-distribution-status` (the "cut" command should return the text before the "=", not the part after that).

The second patch removes the loop, which extracts the distroids from `/etc/armbian-distribution-status` again.
As $filter already contains a list of supported distroids, it should be sufficient to just compare each entry from $filter to $distroid without extracting them again from /etc/armbian-distribution-status.

I am not sure about the additional cut/sed operations in that loop ("cut" removing the part before ";" and "sed" replacing commas with spaces - neither ";" or "," appear in my `/etc/armbian-distribution-status`). If there is a reason for them (I didn't find a syntax definition for `/etc/armbian-distribution-status`), it might be better to omit the second patch (and just include the first patch to correct the field addressed by "cut").

(These fixes do not introduce any new dependencies/modules)

# Documentation Summary
- [ ] **Metadata Included:**  
- [ ] **Document Generated:**  

Both do not apply for these small changes AFAICS.

# Testing Procedure
On Armbian bookworm, execute "armbian-config" and open the "System" menu.
Without the patch, the "Updates" menu is missing.
With the patch, the "Updates" menu is available.

# Checklist

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my code
- [X] I have ensured that my changes do not introduce new warnings or errors
- [X] No new external dependencies are included
- [X] Changes have been tested and verified
(not necessary for these changes)

### DIFF
--- a/tools/modules/system/release_upgrade.sh
+++ b/tools/modules/system/release_upgrade.sh
@@ -26,11 +26,9 @@ release_upgrade(){
 	fi
 
 	local upgrade=$(for j in $filter; do
-		for i in $(grep "^${distroid}" /etc/armbian-distribution-status | cut -d";" -f2 | cut -d"=" -f1 | sed "s/,/ /g"); do
-			if [[ $i == $j ]]; then
-				echo $i
-			fi
-		done
+		if [[ $distroid == $j ]]; then
+			echo $j
+		fi
 	done | tail -1)
 
 	if [[ -z "${upgrade}" ]]; then
--- a/tools/modules/system/release_upgrade.sh
+++ b/tools/modules/system/release_upgrade.sh
@@ -26,7 +26,7 @@ release_upgrade(){
 	fi
 
 	local upgrade=$(for j in $filter; do
-		for i in $(grep "^${distroid}" /etc/armbian-distribution-status | cut -d";" -f2 | cut -d"=" -f2 | sed "s/,/ /g"); do
+		for i in $(grep "^${distroid}" /etc/armbian-distribution-status | cut -d";" -f2 | cut -d"=" -f1 | sed "s/,/ /g"); do
 			if [[ $i == $j ]]; then
 				echo $i
 			fi


igorpecovnik:
> neither ";" or "," appear in my /etc/armbian-distribution-status

We have changed to this below some time ago but BSP package upgrade should carry this file with.

This is mine:
```
cat /etc/armbian-distribution-status
buster=eos;upgrade=bullseye,bookworm,trixie,testing
focal=csc;upgrade=jammy,noble,oracular
bullseye=csc;upgrade=bookworm,trixie,testing
bookworm=supported;upgrade=trixie,testing
jammy=supported;upgrade=noble,oracular
noble=supported;upgrade=oracular
trixie=csc;upgrade=testing
oracular=csc;upgrade=none
sid=csc;upgrade=none
```

torte71:
With that updated /etc/armbian-distribution-status, everything runs fine without my patch, so please ignore my pull request.
Sorry for the noise.

igorpecovnik:
> With that updated /etc/armbian-distribution-status,

Was that update via normal apt update / apt upgrade procedure?

torte71:
No, I just edited that file and replaced its content with the lines you posted.
This happened on an Armbian fork from https://github.com/redrathnure/armbian-mkspi/ , so I think the cause for that outdated/not-updated BSP lies there.

igorpecovnik:
> This happened on an Armbian fork

Aaaah :) Forks are not supported as we have absolutely no knowledge nor control. They are almost certainly outdated, some a little some in years or changed beyond recognition. That work was anyway pulled into Armbian [not long ago](https://github.com/armbian/build/pull/7553) so you don't need to use those forks.

torte71:
That's awesome news!
So I'll happily close this request and start migrating to mainline Armbian. :heart_eyes:

Issue #388, opened 2025-01-23
[Bug]: Kernel header installation fails > pkg_update missing in script


HeyMeco:
### What happened?

`tools/modules/system/install_headers.sh`

can fail to install the kernel headers because it doesn't update the sources prior and then doesn't find the correct version in the repo

### How to reproduce?

1. Don't run apt update for a while
2. Try to install kernel headers via armbian-config

### On which OS are you observing this problem? (for easier diagnosis please provide logs with `armbianmonitor -u`)

Armbian Ubuntu 24.04 Noble

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #389, opened 2025-01-23
Fix Kernel Header installation when apt sources aren't up to date

HeyMeco:
# Description

Add `pkg_update` before trying to install the headers

Issue reference: https://github.com/armbian/configng/issues/388

# Implementation Details

_Provide a detailed description of the implementation. Include the following:_

- [ ] Key changes introduced by this PR

# Documentation Summary

- [ ] **Metadata Included:**  

- [ ] **Document Generated:**  

# Checklist


igorpecovnik (approved):

brentr:
How was this tested?
I'm not seeing how this solves the problem we experienced...
Does installing the kernel-headers somehow also install an updated u-boot bootloader when that kernel update requires it?
@igorpecovnik 
Is there any package whose installation will update the u-boot bootloader version?

igorpecovnik:
> Does installing the kernel-headers somehow also install an updated u-boot bootloader when that kernel update requires it?

No.

> Is there any package whose installation will update the u-boot bootloader version?

You mean like force update of u-boot when package is installed? I think we have a mechanism "force update" of u-boot - package needs to be build this way.

igorpecovnik:
FORCE_UBOOT_UPDATE

brentr:
Sorry. I'm still not understanding.
FORCE_UBOOT_UPDATE and similar flags are part of the Armbian build environment.  Correct?
End users do not use the build environment.

On the RockPI-S (and other) targets:
  apt upgrade
results in an unbootable machine.

I see now that there is a linux-u-boot-rockpi-s-current package, but it is installing the wrong, older v2022.04 bootloader.  Only linux-u-boot-rockpi-s-edge package installs the bootloader the boots the 6.12.9 (current) kernel.  Could this be the problem?

igorpecovnik:
> FORCE_UBOOT_UPDATE default to 'yes' to prevent casual users from ending up with unbootable systems after apt upgrade? 

This will make u-boot auto-update when package is installed. What we can do here is:
1. Adding this variable to the board config
2. When this is in the system, i will build new packages and push them to the repository. This doesn't need to wait for next point release. Can be done "today" with v24.11.x
3. When user will run regular update and upgrade, u-boot package will flash u-boot automatically.

This can be later removed from the sources, or left in so u-boot is always updating.

brentr:
OK. I think I understood now.
It would be great to get the updated uboot packages into the this release.

PR #404, opened 2025-01-31
Enable updates of kernel on all

igorpecovnik:
# Description

Kernel switching condition was wrongly set to only allow this to supported distributions. Which makes no sense. Kernel switching should be enable for all variants.

Also changing keywords to some logical human understandable variants.

# Testing Procedure

```
armbian-config --cmd STABLE
armbian-config --cmd ROLLING
```

# Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code


igorpecovnik:
This was more for the user side part - that menu entry has something logical instead of SY1234 / SY1235.

PR #408, opened 2025-02-02
Fix sanitize

### LINKED ISSUE #324: [Bug]: `sanitize_input` does not work as expected
### What happened?

While testing PR #320 I would get the following error:
```shell
admin@armbian:~/configng$ bin/armbian-config --api pkg_install vim-airline

Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
E: Unable to locate package Invalid argument: vim-airline
```

Upon further testing, any package containing dash (`-`) in the name would produce similar error.

It turns out the error is generated by the `sanitize_input` function which does not allow dashes. However, the pipeline doesn't seem to work as expected, since instead of just reporting the error "Invalid argument: vim-airline," this output is passed further down the pipeline.

In the end the `pkg_install` function tries to install a package called `Invalid argument: vim-airline`, which results in a confusing error.

### How to reproduce?

See above.

### On which host OS are you running the build script and observing this problem?

Ubuntu 24.04 Noble

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

dimitry-ishenko:
# Description

Fixes `sanitize_input` command used with the `--api` and `--cmd` options.

Issue reference:  Closes #324.

# Implementation Details

_Provide a detailed description of the implementation. Include the following:_

- [x] Key changes introduced by this PR

Replaced `sanitize_input` function with simplified version called `sanitize`. Sanitation is only done on the first argument (which is a function name in case of `--api`, or a command name in case of `--cmd`).

_UPDATE:_ The final commit fixes some flawed tests, whose failure was masked by explicit `exit 0` statements in `armbian-config`.

See #324 for more details.

# Testing Procedure

_Describe the tests you ran to verify your changes. Provide relevant details about your test configuration._

```
./bin/armbian-config --cmd DEV-001 # FAILED
./bin/armbian-config --cmd DEV001
./bin/armbian-config --cmd DEV002

./bin/armbian-config --api pkg-install vim-airline # FAILED
./bin/armbian-config --api pkg_install vim-airline
./bin/armbian-config --api pkg_remove vim-airline
```

# Checklist

### DIFF
--- a/bin/armbian-config
+++ b/bin/armbian-config
@@ -83,30 +83,26 @@ case "$1" in
 
     "--cmd")
         INPUTMODE="cmd"
-        shift
-        case "$1" in
+        case "$2" in
             ""|"help")
                 see_cmd_list
             ;;
             *)
-                args=$(sanitize_input "$@")
-                execute_command "$args"
+                cmd=$(sanitize "$2") || die
+                execute_command "$cmd"
             ;;
         esac
     ;;
 
     "--api")
-        shift
-        case "$1" in
+        case "$2" in
             ""|"help")
                 see_use
             ;;
             *)
-                option="$1"
-                shift
-                args=$(sanitize_input "$@")
-                # echo -e "\"$option\" \"$args\""
-                "$option" "$args"
+                fn=$(sanitize "$2") || die
+                shift 2
+                "$fn" "$@"
             ;;
         esac
 	;;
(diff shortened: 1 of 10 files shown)


dimitry-ishenko:
Hmm... not sure why 4 tests fail.

dimitry-ishenko:
OK, so it's really two test cases, not four -- `CON005` and `MON001`. And, the problem is that these test cases are slightly flawed. Fix incoming.

PR #423, opened 2025-02-07
feat: Use deb822 APT sources

blieque:
> This PR and the related PRs will require discussion or RFC before merging, particularly around handling migration of existing installations.

See main PR: https://github.com/armbian/build/pull/7790

- Replace `armbian.list` with `armbian.sources`. This holds the same information in a newer format, deb822.
- Replace HTTP with HTTPS for Armbian repositories.

# Checklist

### DIFF
--- a/tools/modules/system/module_armbian_firmware.sh
+++ b/tools/modules/system/module_armbian_firmware.sh
@@ -259,7 +259,7 @@ function module_armbian_firmware() {
 			local repository=$2
 			local status=$3
 
-			if grep -q 'apt.armbian.com' /etc/apt/sources.list.d/armbian.list; then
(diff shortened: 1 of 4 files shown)


tools/modules/system/module_armbian_firmware.sh
```diff
@@ -229,15 +229,15 @@ function module_armbian_firmware() {
 			local repository=$2
 			local status=$3
 
-			if grep -q 'apt.armbian.com' /etc/apt/sources.list.d/armbian.list; then
+			if grep -q 'apt.armbian.com' /etc/apt/sources.list.d/armbian.sources; then
```
igorpecovnik, line 232: Here we could support both options .list and .sources.

igorpecovnik: Actually backward compatibility here is must have.

Issue #433, opened 2025-02-13
[Bug]: Repository switching doesn't work righrt


igorpecovnik:
### What happened?

sed commands is wrong.

### How to reproduce?

switch stable and beta repo

### On which OS are you observing this problem? (for easier diagnosis please provide logs with `armbianmonitor -u`)

Armbian Ubuntu 24.04 Noble

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #434, opened 2025-02-13
Changing beta and stable repository were not working correct

igorpecovnik:
# Description

- sed command was not working right @blieque 
- we need to make sure /etc/apt/preferences.d/armbian-upgrade-policy is done
- when upgrading, lets specify which packages are needed to not downgrade armbian-config

Closing: https://github.com/armbian/configng/issues/433

# Testing Procedure

- [x] Tested several switches back and forth

# Checklist


blieque:
Thanks for fixing this. My regex required a trailing `/` on the repository URI which should have been optional. I can't quite tell if the URI is supposed to have one or not.

The search regex should have been `/ https?:\/\/[a-z0-9-]+\.armbian\.com\/?( |$)/`.

PR #485, opened 2025-03-21
Derive "branch" and "linuxfamily" from currently installed kernel: fix for kernel hold / unhold

torte71:
# Description

Derive "branch" and "linuxfamily" from currently installed kernel to complete the package names for hold/unhold.

"hold" and "unhold" are calling "show" with empty arguments for "branch" and "linuxfamily", so the actual values have to be calculated within the "show" function (like it is done for the "repository" function, but also including held-back packages in the grep filter).

In its current state, these variables are unset, leading to incomplete package names "linux-image--" and "linux-dtb--", so these packages can't be set to hold/unhold.

# Implementation Details

- [X] Key changes introduced by this PR: Fixes the kernel hold/unhold feature
- [X] Justification for the changes: Bugfix
- [X] Confirmation that no new external dependencies or modules have been introduced

# Documentation Summary

- [?] **Metadata Included:** : Does not apply afaict 

- [?] **Document Generated:** : Does not apply afaict (but `armbian-config --doc` did not emit errors

# Testing Procedure

- [X] Marked/unmarked kernel hold for current and edge kernels, both for stable and rolling package sources.

# Checklist

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my code
- [X] I have ensured that my changes do not introduce new warnings or errors
- [X] No new external dependencies are included
- [X] Changes have been tested and verified
- [?] I have included necessary metadata in the code, including associative arrays

### DIFF
--- a/tools/modules/functions/set_runtime_variables.sh
+++ b/tools/modules/functions/set_runtime_variables.sh
@@ -85,13 +85,26 @@ function set_runtime_variables() {
 
 }
 
-function update_branch_env() {
-	if [[ -z "${BRANCH}" && -f /etc/armbian-release ]]; then
-		BRANCH=$(dpkg -l | grep -E "linux-image" | grep -E "current|vendor|legacy|edge" | awk '{print $2}' | cut -d"-" -f3 | head -1)
+#
+# Retrieve info from currently installed kernel, update /etc/armbian-release if required
+# (after switching kernel, but before a reboot, BRANCH can contain an outdated value)
+#
+function update_kernel_env() {
+	local list_of_installed_kernels=$(dpkg -l | grep '^[hi]i' | grep linux-image | head -1)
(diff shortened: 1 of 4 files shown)


igorpecovnik:
I think it is best to keep this here:
https://github.com/armbian/configng/blob/main/tools/modules/functions/set_runtime_variables.sh#L51-L84

torte71:
Retrieving kernel info only once at script start is not sufficient, as they may change after switching kernels.
Therefore I've moved that into the former update_branch_env() function.

----
Successful tests:

Tests directly after switching kernel (via armbian-config):
- hold/unhold
- install/remove kernel headers
- switch stable/rolling repository

Tests after manually installing custom kernel (via dpkg):
- hold/unhold
- switch stable/rolling repository

igorpecovnik:
> Retrieving kernel info only once at script start is not sufficient, as they may change after switching kernels.
Therefore I've moved that into the former update_branch_env() function.

True.

PR #533, opened 2025-04-17
Add generic "Helper" software help module

Tearran:
This pull request introduces significant refactoring and enhancements to the `module_samba` function and adds a new helper function `_software_help_module`. The changes aim to modularize the code, improve maintainability, and enhance error handling. Below is a summary of the most important changes:

### New functionality and helper function addition:
* Added a new `_software_help_module` function in `tools/modules/functions/helper_help_module.sh` to provide detailed usage instructions and error handling for software modules. This function supports dynamic help generation based on module options.

### Refactoring of `module_samba`:
* Modularized the `module_samba` function in `tools/modules/software/module_samba.sh` by extracting logic into smaller helper functions (`_check_samba_config`, `_set_samba_default`, and `_pipe_service_state`) for better readability and maintainability. [[1]](diffhunk://#diff-bd35f2241870c733836c1cc3358f66145a22cac529dbcf896c3869a157df9618L14-R14) [[2]](diffhunk://#diff-bd35f2241870c733836c1cc3358f66145a22cac529dbcf896c3869a157df9618R23-R25) [[3]](diffhunk://#diff-bd35f2241870c733836c1cc3358f66145a22cac529dbcf896c3869a157df9618L108-R77)
* Simplified the main `case` statement in `module_samba` by delegating help and command handling to `_software_help_module` and other extracted functions.

### Code cleanup:
* Removed redundant comments and unused code from the `module_samba` function to streamline the implementation. [[1]](diffhunk://#diff-bd35f2241870c733836c1cc3358f66145a22cac529dbcf896c3869a157df9618L14-R14) [[2]](diffhunk://#diff-bd35f2241870c733836c1cc3358f66145a22cac529dbcf896c3869a157df9618R23-R25)

### Minor configuration update:
* Added an `enabled` flag in file `2`, possibly to indicate a default state or activation status.

# How was this tested
use as a help message
`./bin/armbian-config --api _software_help_module "samba" module_samba`
or 
` ./bin/armbian-config --api module_samba help`
```
The current Armbian 25.2.3 stable (bookworm) is supported.

Usage: module_samba <command>
Commands: help install remove start stop enable disable configure default status
Available commands:
        install    - install samba
        remove     - remove samba
        start      - start samba
        stop       - stop samba
        enable     - enable samba
        disable    - disable samba
        configure  - configure samba
        default    - default samba
        status     - status samba
```
use for `Invalid command` option
` ./bin/armbian-config --api _software_help_module "cockpit" module_cockpit`
```
[...]

PR #535, opened 2025-04-18
Default module implementation

dimitry-ishenko:
# Description

This module adds default implementation of several functions frequently used by other modules. With this module we can greatly reduce boiler-plate which is currently repeated in many other modules.

For example, to create a simple module that installs midnight commander (mc) now takes only a few lines:
```bash
module_mc()
{
	local module="mc" packages="mc" services=""
	module_default $1 "$module" "$packages" "$services"
}
```
You can test it as follows:
```
admin@uefi-x86:~$ armbian-config --api module_mc help

Usage: module_mc <action> [options...]

Where <action> is one of:
    help        Show this help screen.
    install     Install package(s) for mc.
    remove      Remove package(s) for mc.
    status      Check mc status (installed and/or enabled).

```

If your package actives one or more services, they will be handled by `module_default` as well. So, we can create a module that installs systemd-container:
```bash
module_systemd_container()
{
	local module="systemd_container" packages="systemd-container" services="machines.target"
	module_default $1 "$module" "$packages" "$services"
}
```
And, use it as follows:
```
admin@uefi-x86:~$ armbian-config --api module_systemd_container help

Usage: module_systemd_container <action> [options...]

Where <action> is one of:
    disable     Disable service(s) for systemd_container.
    enable      Enable service(s) for systemd_container.
    help        Show this help screen.
    install     Install package(s) for systemd_container.
    remove      Remove package(s) for systemd_container.
    status      Check systemd_container status (installed and/or enabled).
```

If you need to override default behavior, it can be done as follows:

```
module_systemd_container_install()
{
	module_default_install "$@"

	# additional post-installation actions
	local packages="$1"
	...
}
```

You can likewise override any other action (disable, enable, help, install, remove, status).

You can also add your own actions:

```
module_mc_configure()
{
	local packages="$1" services="$2"
	shift 2
	...
}
```

Issue reference:  #529 #533

# Implementation Details

- [x] Key changes introduced by this PR

This approach not only removes a lot of duplicated boiler-plate, but also completely changes how modules are written. Instead of having one giant monolith case/esac function, which is hard to read and follow, you are a number of small self-contained functions that perform each individual action.

This design also allows us to create higher level helpers that work on top of `module_default` and delegate some of the functions to it. For example, I see there is a number of packages that run inside docker. We can create another module called `module_default_docker` that will perform some of the common functions done by these packages (eg, installing the docker).

Greatly reduces code duplication and boiler-plate. Enforces modularity and encapsulation.

# Documentation Summary

- [x] **Metadata Included:**  

- [ ] **Document Generated:**  

None.

# Testing Procedure

See above.

# Checklist

### DIFF
--- a/tools/modules/functions/module_default.sh
+++ b/tools/modules/functions/module_default.sh
@@ -62,12 +62,12 @@ module_default_help()
 Usage: module_$module <action> [options...]
 
 Where <action> is one of:
-    disable     Disable service(s) for $module.
-    enable      Enable service(s) for $module.
-    help        Show this help screen.
-    install     Install package(s) for $module.
-    remove      Remove package(s) for $module.
-    status      Check $module status (installed and/or enabled)."
+	disable     Disable service(s) for $module.
+	enable      Enable service(s) for $module.
+	help        Show this help screen.
+	install     Install package(s) for $module.
+	remove      Remove package(s) for $module.
+	status      Check $module status (installed and/or enabled)."
 
 	# remove "install" and "remove" lines, if the module doesn't install anything (eg, service-only module)
 	[[ -n "$packages" ]] || text=`grep -Pve " (install|remove) " <<< "$text"`
--- a/tools/modules/functions/module_default.sh
+++ b/tools/modules/functions/module_default.sh
@@ -1,0 +1,106 @@
+# module_default.sh
+
+module_options+=(
+	["module_default,author"]="@dimitry-ishenko"
+	["module_default,desc"]="Default module implementation"
+	["module_default,example"]="disable enable help install remove status"
+	["module_default,feature"]="module_default"
+)
+
+module_default()
+{
+	local action="$1" modules="$2" packages="$3" services="$4"
+	shift 4
+
+	case "$action" in
+		enable)  _module_default_invoke "$action" "$modules" "$services" "$@";;
+		disable) _module_default_invoke "$action" "$modules" "$services" "$@";;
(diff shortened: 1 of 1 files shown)


Tearran:
"The main issue is that Armbian already has a space dedicated to developers. The image-space application is primarily designed for end-users and the image-space itself. It should maintain a balance—rigid development for the core functionality while allowing modules to remain flexible to support alternative approaches.

In my opinion, modules should not rely on the armbian-config library and, where applicable, should avoid using armbian-config as a dependencies altogether. This ensures they can offer unconventional solutions while maintaining simplicity. These modules should focus on generating outputs (stdout) that are fully compatible with armbian-config, without being tied to its internal libraries or structure.

The armbian-config libraries, on the other hand, has focus on parsing stdout from modules, checking OS and hardware, and providing wrappers for UX/UI components."

dimitry-ishenko:
@Tearran I am not sure I understand what you are saying...

Tearran:
Sorry my AI is being long winded as usale. context developer and hobist balance. 

for example as a tinker and hobist. I know `apt-get install` a great app simple to use. and has error handling, can be use directly in module.  Generally speaking, stopping to learning a command that is a shortcut is impractical for time sake to say the least. and would rather spend it learning the build system. That I try to remain ignorant of to keep this balance. 

As a developer that does not like to type the same code over it obviously more efficient to use reusable code.   
And can benefit the project. 

A balance of these need to be met.  Include the libraries you want to include,  but suggest against require hobist and tinkers to adhere to them in the modules. If the knowledge or time to learn is available contributing to the build project is more practical.

PR #537, opened 2025-04-19
Switch adguard home to network host for accurate client ip address logging

aancw:
# Description

This PR changes the Docker network configuration for AdGuard Home from `--net=lsio` to `--network=host`. This change allows AdGuard Home to directly access the host network interfaces. Most importantly, this change fixes the DNS query logging problem by showing the real IP addresses of clients instead of internal Docker network IPs.

Issue reference: https://github.com/AdguardTeam/AdGuardHome/issues/4405
Related documentation: [[Docker networking documentation](https://docs.docker.com/network/host/)](https://docs.docker.com/network/host/)

# Implementation Details
*Provide a detailed description of the implementation. Include the following:*
- [x] Key changes introduced by this PR:
  - Changed Docker network configuration from `--net=lsio` to `--network=host`

- [x] Justification for the changes:
  - **Resolves issue with DNS query logging by showing real client IP addresses instead of internal Docker network IPs (e.g., 172.17.0.x)**

# Documentation Summary
- [ ] **Metadata Included:**  
- [ ] **Document Generated:**  

# Testing Procedure
- [x] Test 1: Verified AdGuard Home container starts successfully with host network
  - Command used: 
    ```bash
    docker run -d \
      --network=host \
      -v "${ADGUARDHOME_BASE}/workdir:/opt/adguardhome/work" \
      -v "${ADGUARDHOME_BASE}/confdir:/opt/adguardhome/conf" \
      --name adguardhome \
      --restart=unless-stopped \
      adguard/adguardhome
    ```
  - Result: Container started successfully without errors
- [x] Test 2: Verified DNS query logging
  - Confirmed that DNS query logs now show the actual client IP addresses (192.168.1.x) 
  - Verified that internal Docker network IPs (172.17.0.x) no longer appear in query logs

**- Before change docker network**
Screenshot 2025-04-19 131357

**- After change docker network**
image

# Checklist
- [x] I have verified that user data in volumes is preserved when migrating from the old configuration
- [ ] I have documented the migration process for existing users

### DIFF
--- a/tools/modules/software/module_adguardhome.sh
+++ b/tools/modules/software/module_adguardhome.sh
@@ -35,7 +35,7 @@ function module_adguardhome () {
 				${module_options["module_adguardhome,feature"]} ${commands[1]}
 			fi
 			docker run -d \
-			--net=lsio \
+			--net=host \
 			-p 53:53/tcp -p 53:53/udp \
 			-p 80:80/tcp -p 443:443/tcp -p 443:443/udp -p 3000:3000/tcp \
 			-p 784:784/udp -p 853:853/udp -p 8853:8853/udp \


igorpecovnik (approved):

fbernier:
@aancw @igorpecovnik  I don't think this is the proper fix and should be reverted. I run a docker bridge network with adguardhome and don't experience the issue described. The root cause of the IPs not being listed is likely something else, such as not having the lsio network created and docker silently falling back to the default bridge network, which does NAT and hides real IPs.

Try running `docker network inspect bridge` and compare it to `docker network inspect lsio`. You should notice that `bridge` has the following option: `"com.docker.network.bridge.enable_ip_masquerade": "true"` which would cause the issue you're seeing. The `lsio` container should not have this option set.

AdguardHome doesn't need access to the host network. Using a docker network provide many benefits here such as network and port isolation.

igorpecovnik:
> I can open a revert PR later if you want.

Go ahead.

Probably same goes for https://github.com/armbian/configng/blob/main/tools/modules/software/module_netdata.sh#L37

@aancw Please verify your network settings.

fbernier:
I made that change to netdata and I think it's correct if we want netdata to be able to monitor the whole network stack of the device it's running on, and not only itself.

igorpecovnik:
Got it, thanks.

aancw:
Thanks for the feedback. I agree with the revert idea. It's better to isolate docker network by default using lsio instead of host.

As far as I remember from my last testing for adguard, it will log docker ip instead of local ip that use adguard dns.

aancw:
I can confirm that switching back to `lsio` does not cause the issue. However, I’m still unsure why it would cause the system to log the Docker subnet instead of the local network.

Here is my `lsio` with adguard home `--net=host`

```
~# docker network inspect lsio
[
    {
        "Name": "lsio",
        "Id": "d05ff8490efccd9c07007f19d992da609b42b67509e784a4a493c77ddbd4771b",
        "Created": "2025-04-09T21:36:11.673235509+07:00",
        "Scope": "local",
        "Driver": "bridge",
        "EnableIPv4": true,
        "EnableIPv6": false,
        "IPAM": {
            "Driver": "default",
            "Options": {},
            "Config": [
                {
                    "Subnet": "172.18.0.0/16",
                    "Gateway": "172.18.0.1"
                }
            ]
        },
        "Internal": false,
        "Attachable": false,
        "Ingress": false,
        "ConfigFrom": {
            "Network": ""
        },
        "ConfigOnly": false,
        "Containers": {
            "b43c089a1251f176a2ac3e03b3972fe745986a728b46f40ffd8e1da881460ae5": {
                "Name": "uptime-kuma",
                "EndpointID": "cd0064c1f94b7af36721d69c930568eebf7a4080bc05b182d0364dfb6f0d6cb4",
                "MacAddress": "fe:af:c3:15:18:c2",
                "IPv4Address": "172.18.0.2/16",
                "IPv6Address": ""
            },
            "b4a7ddd22afd0e9bf71839bc4665c30f63650825050772fa519996600dced9a5": {
                "Name": "stirling-pdf",
                "EndpointID": "03d852efe23811a8ab7297389c49b4d2ac24df7c3bc2d412a629139db8f98181",
                "MacAddress": "fa:ac:e9:e0:b8:07",
                "IPv4Address": "172.18.0.3/16",
                "IPv6Address": ""
            }
        },
        "Options": {},
        "Labels": {}
    }
]
```

Here is my `lsio` with adguard `--net=lsio`
```
[
    {
        "Name": "lsio",
        "Id": "d05ff8490efccd9c07007f19d992da609b42b67509e784a4a493c77ddbd4771b",
        "Created": "2025-04-09T21:36:11.673235509+07:00",
        "Scope": "local",
        "Driver": "bridge",
        "EnableIPv4": true,
        "EnableIPv6": false,
        "IPAM": {
            "Driver": "default",
            "Options": {},
            "Config": [
                {
                    "Subnet": "172.18.0.0/16",
                    "Gateway": "172.18.0.1"
                }
            ]
        },
        "Internal": false,
        "Attachable": false,
        "Ingress": false,
        "ConfigFrom": {
            "Network": ""
        },
        "ConfigOnly": false,
        "Containers": {
            "3af67913450fdeec4c66ff56cdce1400cf2349daada9a6fc79d1af144d811c3f": {
                "Name": "adguardhome",
                "EndpointID": "bae39b312768f8d43b00eac8a21f82f9a95fe25f6f9e3b08e85d8252abcd3173",
                "MacAddress": "2a:d0:9a:0b:b3:f6",
                "IPv4Address": "172.18.0.4/16",
                "IPv6Address": ""
            },
            "b43c089a1251f176a2ac3e03b3972fe745986a728b46f40ffd8e1da881460ae5": {
                "Name": "uptime-kuma",
                "EndpointID": "cd0064c1f94b7af36721d69c930568eebf7a4080bc05b182d0364dfb6f0d6cb4",
                "MacAddress": "fe:af:c3:15:18:c2",
                "IPv4Address": "172.18.0.2/16",
                "IPv6Address": ""
            },
            "b4a7ddd22afd0e9bf71839bc4665c30f63650825050772fa519996600dced9a5": {
                "Name": "stirling-pdf",
                "EndpointID": "03d852efe23811a8ab7297389c49b4d2ac24df7c3bc2d412a629139db8f98181",
                "MacAddress": "fa:ac:e9:e0:b8:07",
                "IPv4Address": "172.18.0.3/16",
                "IPv6Address": ""
            }
        },
        "Options": {},
        "Labels": {}
    }
]
```

image



If the issue is caused by the `lsio` network not being created beforehand, should we create the lsio network interface before installing an app that uses Docker?

fbernier:
> If the issue is caused by the lsio network not being created beforehand, should we create the lsio network interface before installing an app that uses Docker?

Probably yeah. I'm no docker expert and have not investigated this very much but creating a docker container which points to a non-existing network seems to cause issues. I personally switched to a docker-compose setup to manage all my containers on my SBC. Since the network is defined in there, I can't run into such issues. I'm not sure what's the best way forward for armbian-config. I may try to think about it later.

fbernier:
@igorpecovnik thinking about this a little more, I think the root cause of this problem may be that the network name we use throught configng is `lsio`, which may already exist in various forms (with different configurations than a default docker network) if users already have created containers by copy-pasting a command from LinuxServer.io.

Maybe replacing the use of a `lsio` network throughout this repository with another network name such as `armbian` could be an improvement.

igorpecovnik:
> Maybe replacing the use of a lsio network throughout this repository with another network name such as armbian could be an improvement.

Yes, that might be a good idea.

igorpecovnik:
> or can this process be automated through a script?

it's just renaming it:
https://github.com/armbian/configng/blob/main/tools/modules/software/module_docker.sh#L52
and in modules.

PR #665, opened 2025-08-24
Introduce armbian-config config file to store app wide variables

igorpecovnik:
# Description

Purpose of this configuration file is to have a storage for common app wide variables such as DOMAIN name used in SWAG reverse proxy. If domain is enabled, URL to apps is changed from IP to domain name with https enabled.

# Implementation Details

Storing variables in `/etc/armbian-config`

# Testing Procedure

- [x] Store and read while installing SWAG

# Checklist

### DIFF
--- a/tools/modules/functions/set_runtime_variables.sh
+++ b/tools/modules/functions/set_runtime_variables.sh
@@ -40,6 +40,11 @@ function set_runtime_variables() {
 	DIALOG_CANCEL=1
 	DIALOG_ESC=255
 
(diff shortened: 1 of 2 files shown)


igorpecovnik:
I want to improve current operations of some modules. Diving into the v2 code would take me a lot of time ... I barely find time to fix a bugs here and there for what we have now. That's the problem. No doubt about v2 is much better. We need to come up with some strategy how to deal with this (big) change.

PR #741, opened 2026-01-27
modules/software: add HedgeDoc (and fix module_postgres sidecar handling)

igorpecovnik:
Adds **HedgeDoc**, a real-time collaborative markdown editor, as a software module: install / remove / purge / status, backed by its own PostgreSQL container, plus menu entries, help pages, icon and a unit test.

Replaces #718.

## Also fixes three defects in `module_postgres`

They surfaced while getting the test green. All three only affect modules that stand up their **own** database under their own container name — today that is **hedgedoc, immich and netbox** — so the fixes benefit all three.

**1. Port 5432 was published for every container.** `-p 5432:5432` was unconditional even though the container name is a parameter, so no two of those modules can run at once, nor any one of them alongside the stock `postgres` module:

```
Bind for 0.0.0.0:5432 failed: port is already allocated
```

Sidecar databases are reached by container name over the `lsio` network and need no host port. It is now published only for the stock `postgres` container.

**2. `purge` removed the wrong container.** It delegated to `remove` without passing its parameters on; `remove` re-derives the container and image names from them, so it fell back to the defaults. The container being purged survived — and an unrelated one was deleted in its place:

```
Conflict. The container name "/hedgedoc-db" is already in use
```

So purging HedgeDoc's database removed the user's stock `postgres` container, while leaving `hedgedoc-db` behind to collide at the next install.

**3. The wrong image was removed.** HedgeDoc's purge calls named no image, so `rmi` went after the default `tensorchord/pgvecto-rs` and left `postgres:16-alpine` behind. The database identity now lives at function scope so both branches can pass it.

## HedgeDoc no longer ignores a failed database

`module_postgres install` went unchecked, and the readiness gauge is the tail of a pipe — `$?` was the gauge's status, not the wait loop's. A 60-second timeout still printed *"Database is ready for use"* and the install carried on against a database that never came up. Both are checked now.

## Verification

`bash -n` and `shellcheck -s bash -S warning` clean on both modules; `config-assemble.sh -p` succeeds; JSON parses and passes the checks `data-validate-json.yml` runs (no empty `author`, no duplicate ids).

Port and purge behaviour traced through the real call path with the docker helpers stubbed:

| | before | after |
|---|---|---|
| sidecar publishes 5432 | yes | no |
| stock publishes 5432 | yes | yes (unchanged) |
| `purge hedgedoc-db` removes | `postgres` | `hedgedoc-db` |
| image removed | `tensorchord/pgvecto-rs` | `postgres:16-alpine` |

## Note for reviewers

`tests/hedgedoc.conf` follows the existing `unbound.conf` shape, which has no `set -e` — so `testcase()` returns the status of its **last** command (`purge`) and a failing `install` still reports success. That is how the earlier runs on this PR went green while the log contained the port conflict above. Changing it affects ~50 test files, so it is left alone here, but a green run on this suite is not by itself evidence that install worked.


igorpecovnik:
Rebased onto current `main` (the branch was **421 commits** behind, with two merge commits) and squashed to **one commit**: 7 files, +215/−0.

## Review comments

| Comment | Status |
|---|---|
| `exit` instead of `return` terminates the shell | already fixed in `d532756` |
| Duplicate `DATABASE_PASSWORD` generation | already fixed in `d532756` |
| DB password regenerated on reinstall → auth failure | **no longer applies** — the refactor changed the flow: an existing `hedgedoc-db` is now purged and recreated with the new password (module lines 56–65), rather than being skipped. The stale-credential path the comment described is gone. |
| Docs show `https` but `USE_SSL` defaults to `false` | **fixed** — footer now says `http://<your.IP>:3100`, matching `config.runtime.sh`, which already emitted `http://` |

## Three more things found while reviewing

**1. `config.software.json` dropped an unrelated field.** Appending the HedgeDoc entries deleted `"help": "Ghost CMS purge with data folder"` from **`GHOST3`**. Nothing to do with HedgeDoc — restored. That is why the diff is now `+215/−0` instead of `+214/−2`.

**2. `tests/hedgedoc.conf` had no `TESTARCH`.** The branch predates that convention; every test on current `main` sets it. Added `TESTARCH="arm64,amd64"`, matching the module's own `["module_hedgedoc,arch"]="x86-64 arm64"` — without it the test would also be scheduled on armhf/riscv64, where the module is not supported.

**3. The test left its containers behind.** It ended on `status`, so `hedgedoc` and `hedgedoc-db` stayed running on the runner after every execution. Added a trailing `purge`, matching `unbound.conf` and the other module tests.

## Checks

`bash -n` and `shellcheck -s bash -S warning` clean on the module; `config-assemble.sh -p` succeeds; JSON parses and passes the two checks `data-validate-json.yml` runs (no empty `author`, no duplicate ids); `HDOC01`/`HDOC02`/`HDOC03` all resolve.

I did **not** run the test locally — it spins up real `hedgedoc` and `hedgedoc-db` containers, which is CI's job, not my dev machine's. It will run on this PR.

igorpecovnik:
Switched the test to resolute — `tests/hedgedoc.conf` is now `RELEASE="resolute"` (was `noble`). Amended into the same single commit; still 1 commit, 7 files, +215/−0.

CI on this PR will run it as `HedgeDoc install (resolute/amd64)` and `(resolute/arm64)`.

igorpecovnik:
The last run went green but the failure is still there — it just isn't detected. From the `resolute/amd64` log:

```
11:43:50  Starting: hedgedoc-db
11:43:50  Docker run Failed: hedgedoc-db
11:43:50  Bind for 0.0.0.0:5432 failed: port is already allocated
11:44:52  Database 'hedgedoc' is ready for use.
```

The database never started, the 30×2s readiness loop timed out (hence the 62-second gap), and the module announced the database as ready anyway. Two separate defects.

## 1. `module_postgres` publishes 5432 for every container it starts

`-p 5432:5432` was unconditional, but the **container name** is a parameter. Three modules stand up their own database that way — `hedgedoc`, `immich`, `netbox` — so any two of them, or any one of them plus the stock `postgres` module, cannot coexist. On the runner a postgres container from another test already held the port.

Sidecar databases are reached by container name over the `lsio` network and need no host port, so it is now published only for the stock `postgres` container:

```
stock container         -> -p 5432:5432
hedgedoc-db / immich-db -> (no -p)
```

This also unbreaks **immich** and **netbox**, which collide with each other today.

## 2. HedgeDoc ignored the failure

- `module_postgres install`'s return value was not checked.
- The readiness gauge is the **tail of a pipe**, so `$?` was the gauge's status, not the wait loop's. A timeout looked like success.

Both are now checked and return 1 with a message.

## Why CI still passed

Worth flagging separately, because it is not specific to this PR: `testcase()` in the `unbound.conf` style has no `set -e`, so it returns the status of the **last** command — here `purge`, which succeeds regardless. Any failure in `install` is invisible.

The same shape is why the recent ZFS test reports success on noble/resolute while its log contains `Error! Bad return status for module build`. If these tests are meant to gate on install actually working, they need the `set -e` form from `tests/README.md`. I have not changed that here, since it affects ~50 test files and is a separate decision.

**PR is now 1 commit, 8 files, +238/−1.** `bash -n` and `shellcheck -s bash -S warning` clean on both modules; the publish logic verified by stubbing `docker_operation_progress` and inspecting the arguments for stock vs sidecar names.

igorpecovnik:
That one is a different bug, and a nastier one. The module says it is removing the container, but `module_postgres purge` was removing the **wrong** one.

```
"${commands[2]}") # purge
    if ! ${module_options["module_postgres,feature"]} ${commands[1]}; then
```

`purge` delegated to `remove` **with no arguments**. `remove` re-derives the container and image names from `$2..$7`, so with none passed it fell back to the defaults and acted on the stock `postgres` container. So:

- `hedgedoc-db` was never removed → the next install hit `Conflict. The container name "/hedgedoc-db" is already in use`
- and an **unrelated container was deleted in its place** — purging HedgeDoc's database removed the user's stock `postgres` container

Fixed by passing the parameters through.

A second one behind it: HedgeDoc's `purge` calls named no image, so `rmi` targeted the default `tensorchord/pgvecto-rs` and left `postgres:16-alpine` behind. The database identity now lives at function scope so both the install and purge branches can pass it.

Traced with the real call path, stubs on the docker helpers:

| | before | after |
|---|---|---|
| container removed | `postgres` :x: | `hedgedoc-db` :white_check_mark: |
| image removed | `tensorchord/pgvecto-rs:pg14-v0.2.0` :x: | `postgres:16-alpine` :white_check_mark: |
| `hedgedoc-db` left behind | yes :x: | no :white_check_mark: |

`module_hedgedoc purge` now does exactly:

```
rm  -> hedgedoc
rmi -> quay.io/hedgedoc/hedgedoc:latest
rm  -> hedgedoc-db
rmi -> postgres:16-alpine
base_dir remove -> .../hedgedoc-db
base_dir remove -> .../hedgedoc
```

and stock `module_postgres purge` is unchanged (`rm -> postgres`).

That makes three defects in `module_postgres`, all of which only bite modules that stand up their own database under their own name — **hedgedoc, immich and netbox**. The port and purge fixes apply to all three.

PR is 1 commit, 8 files, +248/−3. `bash -n` and `shellcheck -s bash -S warning` clean on both modules.

PR #809, opened 2026-04-08
Fix xmonad desktop session and branding

igorpecovnik:
## Summary
- Use `gnome-session-flashback` for proper xmonad session (provides dbus, polkit, session management)
- Add `.xprofile` for startup apps (wallpaper, xmobar, nm-applet, dunst)
- Add `xinit`, `fonts-ubuntu`, `dmz-cursor-theme`
- Add NM coexistence with systemd-networkd in postinst
- Remove `libghc-xmonad-contrib-dev` (~500MB, not needed for default config)

## Test plan
- [ ] xmonad login via "GNOME Flashback (Xmonad)" session in LightDM
- [ ] Wallpaper, xmobar, nm-applet visible after login
- [ ] Mod4+Shift+Enter opens terminator

### DIFF
--- a/tools/include/branding/postinst/xmonad.sh
+++ b/tools/include/branding/postinst/xmonad.sh
@@ -6,6 +6,20 @@ if [ -d /etc/armbian/lightdm ]; then cp -R /etc/armbian/lightdm /etc/; fi
 # Disable Pulseaudio timer scheduling which does not work with sndhdmi driver
 if [ -f /etc/pulse/default.pa ]; then sed "s/load-module module-udev-detect$/& tsched=0/g" -i  /etc/pulse/default.pa; fi
 
-# set wallpapper to armbian
+# set wallpaper via feh
+mkdir -p /etc/xmonad
+cat > /etc/xmonad/wallpaper.sh <<- 'WALLEOF'
+#!/bin/bash
+feh --bg-scale /usr/share/backgrounds/armbian/armbian03-Dre0x-Minum-dark-3840x2160.jpg
+WALLEOF
+chmod +x /etc/xmonad/wallpaper.sh
 
-
+# Let NetworkManager coexist with systemd-networkd
+if command -v NetworkManager > /dev/null 2>&1; then
+	mkdir -p /etc/NetworkManager/conf.d
+	cat > /etc/NetworkManager/conf.d/10-armbian-unmanaged.conf <<- NMEOF
+	[keyfile]
+	unmanaged-devices=type:ethernet
+	NMEOF
+	systemctl restart NetworkManager 2>/dev/null || true
+fi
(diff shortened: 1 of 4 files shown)


igorpecovnik:
Reviewed. The changes are still worth making, but **this PR cannot be rebased** — it targets files that no longer exist. That is why it shows `CONFLICTING`.

## Blocker: the desktop layout was rewritten under it

| PR touches | On `main` today |
|---|---|
| `tools/include/branding/postinst/xmonad.sh` | :x: gone → `tools/modules/desktops/postinst/xmonad.sh` |
| `tools/include/branding/skel/.xprofile` | :x: gone → `tools/modules/desktops/skel/` |
| `tools/modules/system/module_desktop_packages.sh` | :x: gone → `tools/modules/desktops/yaml/xmonad.yaml` |

Package selection is now YAML with per-release/per-arch overrides, not a bash `case`. The work needs re-applying in the new layout rather than a merge.

## The intent has not landed — still worth doing

- `postinst/xmonad.sh` on `main` still ends at the bare `# set wallpapper to armbian` placeholder (typo included). No wallpaper, no NM handling.
- No `.xprofile` anywhere in `desktops/skel/`.
- `xmonad.yaml` minimal tier still lists `xterm` and none of `gnome-session-flashback`, `xinit`, `dmz-cursor-theme`, `fonts-ubuntu`.

One bullet is already moot: **`libghc-xmonad-contrib-dev` is not in `xmonad.yaml` at all**, so the ~500 MB saving has already happened elsewhere.

## Packages check out

All four additions verified against `xmonad.yaml`'s full declared matrix — 7 releases × up to 5 arches, **28 combinations** including `armhf`, `riscv64` and `loong64` — using the repo's own `audit.py` availability checker:

`gnome-session-flashback` · `xinit` · `dmz-cursor-theme` · `fonts-ubuntu` — **available everywhere, no gaps.** No per-release or per-arch override needed.

## The open review comment is still valid

`.xprofile` in `desktops/skel/` **would** leak into every desktop. `module_desktop_branding.sh:48` copies the shared skel wholesale for whichever DE is installing:

```bash
if [[ -d "$desktop_dir/skel" ]]; then
    cp -R "$desktop_dir/skel/." /etc/skel/
fi
```

There is one shared `skel/`, not a per-DE one — so `xmobar &`, `nm-applet &` and `dunst &` would start under XFCE, GNOME and KDE too. Either guard on the session as suggested, or drop it in from the xmonad postinst instead of skel.

## Two more things to settle before re-doing this

**1. The session rename likely breaks autologin.** `packages/bsp/common/usr/lib/armbian/armbian-firstlogin:1097-1098` forces the session by name:

```bash
[[ -x $(command -v xmonad) ]] && sed -i "s/user-session.*/user-session=xmonad/" ...
```

If the session becomes *"GNOME Flashback (Xmonad)"*, `user-session=xmonad` no longer names an existing session and first login may not land where intended. `xmonad.yaml` also still declares `display_manager: lightdm`. Worth checking what `.desktop` name flashback actually installs and updating both.

**2. `unmanaged-devices=type:ethernet` is a system-wide change.** Handing *all* ethernet to systemd-networkd from a window-manager postinst affects the whole machine, not just the xmonad session, and it is written unconditionally whenever NetworkManager is present. Worth confirming that is intended for every xmonad install.

FWIW `xmonad.yaml` carries `status: unsupported`, so none of this is on a release-critical path.

PR #845, opened 2026-04-13
manage_dtoverlays: atomic boot-config writes + safety hardening

igorpecovnik:
## Summary

Boot config edits are reboot-critical on an SBC — a sed-in-place that fails partway leaves the board unbootable. Move the write to temp + atomic `mv` with a `.bak` preserved from the previous content, matching the pattern `module_desktop_repo` uses for `sources.list.d` and `preferences.d`.

While in here, the rest of the high-value review items:

- `exit 0` / `exit 1` inside the function → `return 0` / `return 1` (`exit` kills the whole `armbian-config` invocation)
- Anchored `/^overlay_prefix=/` in the non-Pi branch — the previous unanchored `awk` would also match keys like `something_overlay_prefix_FOO=`
- Consistent quoting of `$overlayconf` / `$overlay` / `$overlaydir` across `grep`/`sed`/`find`
- `grep -wF --` for overlay-name lookups so dots and hyphens in names don't get interpreted as regex
- `set -o pipefail` on the `find | awk | sort` discovery pipelines so a transient `find` failure doesn't silently mask all overlays
- Match `^overlays=` (not `^overlays`) so we don't accidentally hit `overlay_prefix=`
- `help` subcommand prints usage and returns 0 (matches the `module_desktop_repo` / `module_desktops` convention); empty arg still drops into the TUI since that's this module's primary mode
- `module_options.arch` gains `riscv64` — overlays are increasingly relevant on RISC-V SBCs we now build for
- Locals: `scenario`, `available_overlays`, `builtin_overlays`, `exit_status`, `selection`, `ovs`/`ov`, `candidate`, `status`, `overlay` are now `local` (no parent-shell leakage)
- Dropped dead `j=0`
- Factored the `LINUXFAMILY=="bcm2711"` branching into a single `is_pi` local at the top so the Pi-vs-Armbian split is explicit, instead of three scattered checks

## Test plan

The write helper (`_dtoverlays_write_config`) was smoke-tested against four scenarios with byte-level comparison. All produce byte-identical output to the previous function, plus a `.bak`:

| scenario | input | result |
|---|---|---|
| armbianEnv: existing `overlays=` line | `overlays=foo bar` | replaced with new selection; `.bak` keeps old |
| armbianEnv: no `overlays=` line | (none) | appended at EOF |
| Pi config: no `# Armbian config` block | (none) | block created at EOF |
| Pi config: existing `# Armbian config` block | `dtoverlay=old1`, `old2` | block replaced from marker to EOF |

Items intentionally left for a follow-up PR (cosmetic, not in scope here): direct call to `reboot` vs `systemctl reboot`, hard-coded boot paths, and the bigger refactor splitting discovery/apply for CLI mode.

- [ ] `armbian-config --api manage_dtoverlays help` prints usage and exits 0 without launching the TUI
- [ ] On a real Armbian board, toggling overlays leaves `/boot/armbianEnv.txt.bak` containing the previous content and `/boot/armbianEnv.txt` containing the new selection
- [ ] On a Raspberry Pi, the same applies to `/boot/firmware/config.txt[.bak]`
- [ ] On a `rk3588` vendor-kernel board, builtin (non-prefixed) overlays still appear in the picker

### DIFF
--- a/tools/modules/system/manage_dtoverlays.sh
+++ b/tools/modules/system/manage_dtoverlays.sh
@@ -9,56 +9,89 @@ module_options+=(
 ["manage_dtoverlays,status"]="Active"
 ["manage_dtoverlays,group"]="Kernel"
 ["manage_dtoverlays,port"]=""
-["manage_dtoverlays,arch"]="aarch64 armhf"
+["manage_dtoverlays,arch"]="aarch64 armhf riscv64"
 )
 #
 # @description Enable/disable device tree overlays
 #
-function manage_dtoverlays () {
-	# check if user agree to enter this area
+# Usage: manage_dtoverlays
+#        manage_dtoverlays help
+#
+# Edits /boot/armbianEnv.txt (or /boot/firmware/config.txt on
+# Raspberry Pi). The file is rewritten via temp + atomic mv and
+# the previous content is preserved as <name>.bak — on an SBC a
+# corrupted boot config is a brick, so the rewrite must be all-or-
+# nothing.
+#
+function manage_dtoverlays() {
+	local arg="${1:-}"
+
+	case "$arg" in
+		help)
+			echo "Usage: manage_dtoverlays"
+			echo "       manage_dtoverlays help"
+			echo ""
+			echo "Interactive TUI for enabling/disabling device tree overlays."
+			echo "Edits /boot/armbianEnv.txt or /boot/firmware/config.txt."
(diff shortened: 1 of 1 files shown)


igorpecovnik:
Folded into #846 — that branch already includes this commit (`15a63ade`) as the base of its 3-commit stack. Closing to consolidate review.

PR #849, opened 2026-04-13
desktops: drop plucky and questing release blocks

igorpecovnik:
## Summary

`plucky` (Ubuntu 25.04) is **eos** as of 2026-01-25 — see [armbian/build#9657](https://github.com/armbian/build/pull/9657) where we flipped its support flag. The audit script already excludes eos releases from "missing", but it does not auto-remove existing entries; those just sit as drift across every DE YAML.

`questing` (Ubuntu 25.10) is a 9-month interim release with about **3 months of life left** (EOL ~2026-07). Rare on real boards and not worth the per-release maintenance overhead.

This sweep removes both from every desktop YAML.

### Dropped

| file | what |
|---|---|
| `common.yaml` | browser map (plucky, questing); tier_overrides.mid loupe blocks (plucky, questing); tier_overrides.full thunderbird blocks (plucky, questing) |
| `xfce.yaml`, `gnome.yaml`, `kde-plasma.yaml`, `kde-neon.yaml`, `mate.yaml`, `cinnamon.yaml`, `i3-wm.yaml`, `xmonad.yaml`, `enlightenment.yaml`, `budgie.yaml`, `deepin.yaml` | `plucky:` and `questing:` release blocks |

`bianbu.yaml` was already noble+resolute only — nothing to drop there.

Comments mentioning plucky/questing as historical context (e.g. "pavumeter dropped in plucky") are kept as reference.

### Knock-on

- [armbian/os#444](https://github.com/armbian/os/pull/444) drops the questing aptly configs (`chromium-aarch64-questing.conf`, `firefox-questing.conf`, `thunderbird-questing.conf`) since this PR removes their consumer.

### Stack

Built on top of [armbian/configng#848](https://github.com/armbian/configng/pull/848) (apt.armbian.com browser routing). Will rebase cleanly when #848 merges.

## Test plan

- [x] `parse_desktop_yaml.py` — `DESKTOP_SUPPORTED="yes"` for every kept release × arch on xfce; `"no"` for plucky/questing
- [x] All 12 DE YAML files still parse (sanity ran across the matrix)
- [ ] Audit workflow on next run does not report plucky as a missing release (it's eos, already excluded)
- [ ] No remaining build/CI matrix references to plucky/questing in configng

PR #850, opened 2026-04-13
desktops: install runtime apt pin so apt.armbian.com .debs win

igorpecovnik:
## Summary

The desktop YAML lists `chromium` / `firefox` / `thunderbird` / `chrome` / `code` / `edge` expecting apt.armbian.com to host them as real .debs that should win over Ubuntu's snap-transitional packages. For that to actually happen, an apt pin must force `o=Armbian` to priority **1001** and push `o=Ubuntu` down to **50**.

Today that pin lives in the BSP at [packages/bsp/common/etc/apt/preferences.d/armbian](https://github.com/armbian/build/blob/main/packages/bsp/common/etc/apt/preferences.d/armbian). That file is a **dpkg conffile** — once a user has any version on disk (even the May-2023 → Feb-2026 era one which was shipped fully commented out), `armbian-bsp-cli` upgrades preserve the local copy instead of installing the new one. Result on those systems: the pin is inert and the snap-shim wins. Exactly the symptom we just debugged on a real test box.

## Fix

Move pin ownership to armbian-config and write to a **distinct path** so dpkg's conffile preservation is moot:

- `_module_desktops_write_apt_pin()` writes `/etc/apt/preferences.d/armbian-desktops` via temp + atomic `mv`. Idempotent — runs on every `install`, always reflects the current configng version.
- Wired into the `install` subcommand **before** `pkg_update` / `pkg_install` so apt resolves the desktop package list with the right priority from the first invocation.
- Non-fatal: pin write failure warns and continues — better than aborting a desktop install over a preferences file.

Pin block emitted:

```
Package: chromium chromium-* firefox firefox-esr firefox-l10n-* thunderbird thunderbird-l10n-* google-chrome-stable code microsoft-edge-stable
Pin: release o=Armbian
Pin-Priority: 1001

Package: chromium chromium-* firefox firefox-esr firefox-l10n-* thunderbird thunderbird-l10n-* google-chrome-stable
Pin: release o=Ubuntu
Pin-Priority: 50
```

### Why 1001, not 990

Ubuntu's snap-transitional packages have a higher epoch than Armbian's real .debs. Priority 990 only permits upgrades; 1001 also permits the downgrade required to swap an existing snap-shim install for the real .deb. The build-repo HEAD pin already uses 1001 — this matches.

### Stack

Built on top of:
- [armbian/configng#848](https://github.com/armbian/configng/pull/848) — apt.armbian.com browser routing
- [armbian/configng#849](https://github.com/armbian/configng/pull/849) — drop plucky/questing

Will rebase cleanly when those land.

### Build-repo follow-up

Newly-built images still install desktop packages via the build framework's own apt path, so they'd get the snap-shim until either:

1. The build invokes `module_desktops install` for desktop images (long-term — single source of truth), or
2. The build also writes the same pin file from its desktop hook (short-term bridge)

Separate PR. The BSP pin file at `packages/bsp/common/etc/apt/preferences.d/armbian` should also be deleted (or kept minimal) once the runtime pin is the source of truth.

## Test plan

- [x] Helper renders the expected pin file content (smoke-tested in isolation)
- [x] Bash syntax check of the modified module
- [ ] On a real Armbian noble system: `armbian-config --api module_desktops install de=xfce tier=full` produces `/etc/apt/preferences.d/armbian-desktops` with both pin blocks, and `apt-cache policy thunderbird` shows Armbian = 1001, Ubuntu = 50, snap-shim NOT installed
- [ ] On a system that already had the snap-shim installed: install path downgrades it to the apt.armbian.com real .deb (1001 forces the swap)
- [ ] Pin write failure (e.g. read-only `/etc`) warns and the install still proceeds

### DIFF
--- a/tools/modules/desktops/module_desktops.sh
+++ b/tools/modules/desktops/module_desktops.sh
@@ -162,8 +162,15 @@ function module_desktops() {
 				echo "Warning: '${de}' is not supported on ${DISTROID}/$(dpkg --print-architecture)" >&2
 			fi
 
-			# suppress interactive prompts
-			echo "encfs encfs/security-information boolean true" | debconf-set-selections 2>/dev/null || true
+			# Suppress interactive prompts. The `code` (Microsoft VSCode)
+			# postinst asks whether to add the Microsoft apt repository
+			# — say no, because apt.armbian.com already hosts code and
+			# adding the parallel Microsoft source would race against
+			# our pin on every apt-get update.
+			debconf-set-selections 2>/dev/null <<- 'EOF' || true
+			encfs encfs/security-information boolean true
+			code code/add-microsoft-repo boolean false
+			EOF
 
 			# set up custom repo if needed
 			if ! module_desktop_repo "$de"; then
--- a/tools/modules/desktops/module_desktops.sh
+++ b/tools/modules/desktops/module_desktops.sh
@@ -30,6 +30,78 @@ _desktop_in_container() {
 }
 
 #
+# Write the apt pin that forces apt.armbian.com .debs to win over
+# Ubuntu's snap-transitional packages for desktop apps. Idempotent —
+# writes via temp + atomic mv so a partial write never leaves apt with
+# an unparseable preferences file.
+#
+# Lives at /etc/apt/preferences.d/armbian-desktops (note: NOT the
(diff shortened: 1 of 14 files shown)


igorpecovnik:
@copilot resolve the merge conflicts in this pull request

PR #853, opened 2026-04-14
desktops: drop libu2f-udev from resolute (no installation candidate)

igorpecovnik:
## Summary

`libu2f-udev` was a transitional Ubuntu/Debian package that shipped udev rules for U2F security keys. The rules have since been absorbed into `udev` itself (and `libfido2-1` for newer FIDO2 flows), so resolute (Ubuntu 26.04 LTS) drops the transitional package from the archive entirely.

Reported symptom on a real resolute desktop install:

```
E: Package 'libu2f-udev' has no installation candidate
```

## Fix

Strip `- libu2f-udev` from each DE YAML's `resolute:` release block — 8 files: `cinnamon`, `enlightenment`, `gnome`, `i3-wm`, `kde-plasma`, `mate`, `xfce`, `xmonad`.

Other release blocks keep it — bookworm/trixie/noble/jammy/forky/sid still ship the transitional package and the desktop install path on those releases continues to install it. Bianbu has a resolute block but never carried libu2f-udev, so no change there.

## Test plan

- [x] xfce.yaml `resolute:` block no longer references libu2f-udev (other release blocks still do)
- [ ] On real resolute board: `armbian-config --api module_desktops install de=xfce tier=full` no longer hits the missing-candidate error
- [ ] U2F security keys still work after install (udev rules now provided by `udev` upstream — verify with `udevadm info` or by plugging in a Yubikey)

### DIFF
--- a/tests/BUDG01.conf
+++ b/tests/BUDG01.conf
@@ -1,5 +1,5 @@
 ENABLED=false
-RELEASE="trixie:noble"
+RELEASE="trixie:noble:resolute"
 TESTNAME="Budgie desktop"
 
 testcase() {(
--- a/tests/CINM01.conf
+++ b/tests/CINM01.conf
@@ -1,5 +1,5 @@
 ENABLED=true
-RELEASE="bookworm:trixie:noble"
+RELEASE="bookworm:trixie:noble:resolute"
 TESTNAME="Cinnamon desktop"
 
 testcase() {(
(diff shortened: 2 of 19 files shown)


EvilOlaf:
https://packages.debian.org/en/trixie/libu2f-udev

> This package is not necessary anymore, and can be safely removed: since udev v244, U2F devices are autodetected without needing 3rd party udev rules. 

So should be good to drop

PR #886, opened 2026-04-21
desktops: enable panthor-gpu DT overlay on rk3588 vendor-kernel desktops

igorpecovnik:
## Summary

Ports armbian/build's `extensions/mesa-vpu.sh` `extension_prepare_config__3d` hook into `module_desktops install` so the `panthor-gpu` DT overlay is enabled at the desktop-install layer, not only at image-build time.

Mesa's **panthor** driver needs the `panthor-gpu` DT overlay to claim the rk3588 Mali-G610 via the vendor kernel's kbase interface. Without it, GL / Vulkan / GBM fall back to llvmpipe software rendering. Previously the build-side extension appended `panthor-gpu` to `DEFAULT_OVERLAYS` at `extension_prepare_config` time — fine for image-built desktops, but desktops installed on top of a minimal image via `armbian-config` never got the overlay.

## Change

New helper `_module_desktops_add_3d_overlay` in `module_desktops.sh` with the same gating upstream has:

- `BOARDFAMILY` in `{rockchip-rk3588, rk35xx}`
- `BRANCH == vendor`
- release **not** in `{bookworm, bullseye, buster, focal, jammy}` (panthor kernel bits unusable on those)
- `tier != minimal` (overlay is only useful with a Mesa/Vulkan stack that mid/full pulls in)
- `DESKTOP_ENVIRONMENT` **not** in `{xfce, i3-wm}` (X11-only in our defaults, no GBM path)

All conditions match → delegate to the existing `module_devicetree_overlays install overlays=panthor-gpu`. That module already owns:

- atomic `/boot/armbianEnv.txt` rewrite (temp + mv, `.bak` preserved)
- validation against the discovered `.dtbo` set
- idempotence (silently no-op if `panthor-gpu` already enabled)

No reimplementation needed.

## Callsite

Wired into the install pipeline right after `_module_desktops_configure_networking`, so both build-mode (`armbian-config --api module_desktops install mode=build`) and runtime installs converge on the same overlay set image-built desktops already have.

## Environment

`BOARDFAMILY` and `BRANCH` are sourced from `/etc/armbian-release` at configng init time by `module_env_init.sh` — present in the chroot during `mode=build` because `armbian-base-files` (which ships the file) is installed before `module_desktops` on every build path.

## Companion PR

Pairs with armbian/build#9683 commit `deb764dc6`, which removes `extension_prepare_config__3d` from `extensions/mesa-vpu.sh` (and the `post_armbian_repo_customize_image__browser` hook — also redundant with configng's `browser` virtual token, and the cause of the chromium master_preferences dpkg prompt that kicked off this refactor).

## Test plan

- [x] `bash -n module_desktops.sh` passes.
- [x] Helper is a no-op on non-matching `BOARDFAMILY` / `BRANCH` / release / tier / DE (gating mirrors mesa-vpu exactly).
- [ ] Image build of noble/rk3588/vendor/gnome/mid ends with `panthor-gpu` in `/boot/armbianEnv.txt` `overlays=` line.
- [ ] Runtime `armbian-config --api module_desktops install de=gnome tier=mid` on a noble/rk3588/vendor minimal image writes `panthor-gpu` into `/boot/armbianEnv.txt` idempotently and keeps a `.bak`.
- [ ] Same call on an rk3566 or x86 board: helper early-returns silently, no overlay write.

### DIFF
--- a/tools/modules/desktops/module_desktops.sh
+++ b/tools/modules/desktops/module_desktops.sh
@@ -102,6 +102,86 @@ function _module_desktops_write_apt_pin() {
 }
 
 #
+# Enable the panthor-gpu DT overlay on Rockchip RK3588-family boards
+# running the vendor kernel, when a Wayland-capable desktop is being
+# installed. Mirrors the old extensions/network/… no, wait — mirrors
+# armbian/build's extensions/mesa-vpu.sh 'extension_prepare_config__3d'
+# hook so that install-on-minimal converges on the same overlay set
+# an image-built desktop would have.
+#
+# panthor-gpu is the Mesa panthor-kbase GPU driver overlay: required
+# for hardware-accelerated GL / Vulkan / GBM on rk3588 under Mesa +
+# vendor kernel, unused (and ignored) on everything else.
+#
+# Gating mirrors the upstream hook:
+#   - Skip on pre-bookworm / bookworm / jammy / older releases —
+#     panthor kernel bits didn't land there in a usable shape.
+#   - Skip on tier=minimal — the overlay is only useful with a
+#     compositor stack (Mesa/Vulkan loaders) that mid/full pulls in.
+#   - Skip on xfce / i3-wm — these are X11-only in our default
+#     configuration and don't benefit from a GBM GPU path.
+#   - Only fire on BOARDFAMILY rockchip-rk3588 / rk35xx and
+#     BRANCH=vendor.
+#
+# BOARDFAMILY + BRANCH are globals set at configng init time by
+# module_env_init.sh (which sources /etc/armbian-release); present
+# in the chroot under mode=build because armbian-base-files is
+# installed before module_desktops. Delegates the actual
+# armbianEnv.txt write to module_devicetree_overlays (atomic
+# temp+mv, .bak preserved, validates name against the discovered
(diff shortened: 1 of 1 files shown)


HeyMeco:
@igorpecovnik what would be an approach for when we want to have Mali GPU driver instead of Panthor on the final build. Since the mesa-vpu extension is not in build anymore and this seems to be a default how can we overwrite this with an extension in build?

EvilOlaf:
Tried to re-use the function name? Usually if you use a function name within an extension it will override the one in the framework. Somethin like

```bash
_module_desktops_add_3d_overlay() {
    :
}
```

HeyMeco:
@EvilOlaf I'm currently figuring out the architectural approach so I haven’t tried it yet. Can you link me (if there are) any examples where build overwrites a function from configng?

EvilOlaf:
oh, that's within configng. Then, I'm afraid, no, don't know an example. Well honestly I don't know one from framework  by heart either, would need to search for one.

HeyMeco:
Yeah not really a fan of the move of this to configng. At least the panthor overlay aspect should probably stay in build as an extension while the packages make sense to be in the desktop config here.

PR #907, opened 2026-05-09
runtime: route menu URLs through get_service_url for SWAG-aware modules

igorpecovnik:
## Summary

[`config.runtime.sh`](tools/modules/runtime/config.runtime.sh) writes the per-service info shown on each menu item's right column. Lines using a hardcoded `http://$DISPLAY_URL:${module_options[…,port]}` template kept showing the direct host:port even when the matching service was fronted by SWAG, so users saw e.g. `Bazarr remove (http://test.armbian.com:6767)` instead of the SWAG subfolder URL.

The `get_service_url` helper that powers `TRA002` (transmission), `IMM002` (immich), and `NBOX02` (netbox) already does the right thing:
- no SWAG running → `http://$DISPLAY_URL:port`
- SWAG running but service-conf not enabled → `http://$DISPLAY_URL:port`
- SWAG running and service-conf enabled → `https://$DISPLAY_URL/<service>`

This PR routes the remaining menu lines whose service has a matching stock proxy-conf in [`linuxserver/reverse-proxy-confs:master`](https://github.com/linuxserver/reverse-proxy-confs) through that helper.

Pairs naturally with #906 (which `enables` the proxy-conf at install time): once a user installs a wired module on a host that runs SWAG, the menu reflects the SWAG path on the next refresh.

## Switched (23 lines)

bazarr, deluge, domoticz, duplicati, embyserver → `emby`, filebrowser, ghost (path `/ghost`), grafana, homepage, jellyfin, lidarr, medusa, netdata, phpmyadmin, pi-hole → `pihole` (path `/admin`), plexmediaserver → `plex`, prowlarr, qbittorrent, radarr, readarr, sabnzbd, sonarr, syncthing.

## Intentionally left untouched

- `NCT002` (nextcloud) and `CPT002` (cockpit): the direct URL is HTTPS on a self-signed port; `get_service_url`'s fallback returns `http://…` on hosts without SWAG, which would break those panels. Worth a follow-up that extends the helper with a fallback-protocol argument.
- `POR002` (portainer): already has its own explicit SWAG-aware branch right above this section; no change needed.
- Modules whose service has no stock SWAG proxy-conf (actualbudget, adguardhome, octoprint, openhab, …): fallback would equal current behaviour but offer no upside, and routing through the helper would drag in a per-row exec into the SWAG container for nothing. Kept on the direct template.

## Test plan

- [ ] On a host **without** SWAG: menu items for the 23 switched services still show `http://<host>:<port>` exactly as before.
- [ ] On a host **with** SWAG and the matching proxy-conf enabled (e.g. via #906): `Bazarr remove (https://my.home.org/bazarr)` instead of `(http://my.home.org:6767)`.
- [ ] `pi-hole` row still ends with `/admin`; `ghost` row still ends with `/ghost`.
- [ ] No regression on the three rows that already used the helper (transmission, immich, netbox).


igorpecovnik:
Superseded — squashed into #906 alongside the per-module wiring.

PR #958, opened 2026-08-03
add lxqt as a desktop option

Awire9966:
Hello! 
This PR is my fix to Issue: [#10306](https://github.com/armbian/build/issues/10306)
It adds the LXQT desktop as an option.
I added a file: https://github.com/Awire9966/configng/blob/main/tools/modules/desktops/yaml/lxqt.yaml

### DIFF
--- a/tests/LXQT01.conf
+++ b/tests/LXQT01.conf
@@ -1,0 +1,10 @@
+ENABLED=true
+RELEASE="bookworm:trixie:noble:resolute"
+TESTARCH="arm64,amd64"
+TESTNAME="LXQt desktop"
+
+testcase() {(
+	set -e
+	./bin/armbian-config --api module_desktops install de=lxqt tier=full
+	./bin/armbian-config --api module_desktops remove  de=lxqt
+)}
--- a/tests/LXQT01.conf
+++ b/tests/LXQT01.conf
@@ -1,5 +1,5 @@
 ENABLED=true
-RELEASE="bookworm:trixie:noble:resolute"
+RELEASE="trixie:resolute"
 TESTARCH="arm64,amd64"
 TESTNAME="LXQt desktop"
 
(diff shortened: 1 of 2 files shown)


igorpecovnik:
Thanks for adding LXQt, @Awire9966! I've pushed a commit to complete it to a working state (maintainer edit) — kept you as author via `Co-authored-by`.

The main blocker: the YAML had no `releases:` block, and the parser gates availability on `arch in supported_archs AND release in releases`, so with an empty `releases` the desktop resolved to `DESKTOP_AVAILABLE="no"` on every release/arch — it could never be selected or installed.

What the follow-up commit does (modelled on `kde-plasma.yaml`, the closest Qt desktop, also on sddm):

- **Adds a full `releases:` block** (bookworm/trixie/noble/forky/sid/jammy/resolute) with per-release architectures and the pulseaudio→pipewire swap on trixie/forky/sid — this is what makes it installable.
- **Drops the hardcoded `firefox-esr`** in `mid`: `common.yaml` already contributes a per-arch `browser` token (tiers are additive), and `firefox-esr` isn't in the Ubuntu archive, so it would have broken installs on noble/jammy/resolute.
- **Drops `snap`** from `full` (not a real package name — `snapd` ships the `snap` binary; Armbian also avoids snap).
- **Trims packages already provided by common's mid/full** (bluez/blueman/thunderbird/libreoffice/vlc…) and swaps the GTK editor/archiver for LXQt's `featherpad`/`lxqt-archiver`.
- **Drops the invalid `lxqt-branding`** package and uses **openbox** (the LXQt WM) instead of `xfwm4`; normalises indentation.

Validated with `parse_desktop_yaml.py` across releases/arches/tiers: `DESKTOP_AVAILABLE=yes` where expected, the browser resolves from the shared per-arch table (chromium/chrome, never firefox-esr on Ubuntu), no `snap`, and `lxqt` now lists alongside `xfce`/`kde-plasma`. `display_manager: sddm` was a good call — the framework auto-installs the DM package.

PR #963, opened 2026-08-05
armbian-install: make eMMC installs boot on RK3588 (u-boot offset + GPT table)

igorpecovnik:
Makes the new install engine produce a **bootable eMMC** image. Found and fixed against **Banana Pi M7 (RK3588)** with serial-console confirmation — two independent bugs, both required for boot.

## Bug 1 — first partition overlapped u-boot

`install_apply_partitions` started the first partition at a fixed **1 MiB** for every mode. In `emmc` mode u-boot is dd'd to raw sectors of the same device (`u-boot.itb` at sector 16384 = **8 MiB**), so a 1 MiB partition overlaps it → filesystem and bootloader clobber each other.

**Fix:** thread a `start=` offset through the plan — `start=16` (MiB) for `emmc` (matching the classic `FIRSTSECTOR=32768`), `start=1` otherwise. `install_apply_partitions` honours it.

## Bug 2 — MBR table the board's u-boot can't read

With Bug 1 fixed, u-boot is present and clear, but the board still wouldn't boot. Serial showed the **Rockchip vendor u-boot 2017.09** reading the eMMC as **GPT only**:

```
part_get_info_efi: *** ERROR: Invalid GPT ***
part_get_info_efi: *** ERROR: Invalid Backup GPT ***     (× hundreds)
android_image_load_by_partname: Can't find part: boot
FIT: No boot partition ... Unknown command 'bootrkp'
```

`install_table_type` picked `msdos` for a small non-UEFI disk; this vendor u-boot parses GPT only, so it never finds the kernel and falls through Android→PXE→FIT.

**Fix:** replicate the running image's partition-table type (what the classic installer does), rather than deciding purely by capacity. `install_source_table_type` detects the source root disk's `PTTYPE`; `install_run_scenario` passes it to the planner for `emmc`/`sd`. Forcing GPT universally would be wrong — Allwinner needs MBR because its SPL at sector 16 (8 KiB) overlaps a GPT partition-entry array — so this **inherits** rather than forces. Hard requirements (UEFI, >2 TiB, 4Kn) still upgrade to GPT.

## Tests

`plan.bats` (23) + `detect.bats` (7) + `bootconfig.bats` (26) green locally; `integration_loopback` skips without root (runs in CI). New cases cover the 16 MiB emmc offset, the table-type preference (honoured / overridden by hard requirements), and emmc gpt-source→gpt / msdos-source→msdos.

## Hardware validation

Root-caused via UART on the M7 (RK3588 vendor u-boot). Needs a confirming eMMC install + boot on that board once merged.

### DIFF
--- a/tests/bats/detect.bats
+++ b/tests/bats/detect.bats
@@ -69,6 +69,32 @@ setup() {
 	[[ "$output" != *"rpmb"* ]]
 }
 
+@test "source table: selects the /boot device's type, not /'s, when they differ" {
+	# SD /boot (gpt) + NVMe root (msdos): u-boot reads /boot, so the target must
+	# replicate the /boot device's table (gpt), not the root device's.
+	findmnt() { case "$*" in *' /boot'*) echo /dev/mmcblk0p1 ;; *) echo /dev/nvme0n1p1 ;; esac; }
+	lsblk()   {
+		local a="$*"
+		if [[ "$a" == *PKNAME* ]]; then
+			[[ "$a" == *mmcblk0p1* ]] && echo mmcblk0
+			[[ "$a" == *nvme0n1p1* ]] && echo nvme0n1
+		elif [[ "$a" == *PTTYPE* ]]; then
+			[[ "$a" == *mmcblk0* ]] && echo gpt
+			[[ "$a" == *nvme0n1* ]] && echo dos
+		fi
+	}
+	run install_source_table_type
+	[ "$output" = "gpt" ]
+}
+
+@test "source table: falls back to / when /boot is not a separate mount" {
+	# /boot is a dir on / (no mount) -> use the root device's table.
+	findmnt() { case "$*" in *' /boot'*) echo "" ;; *) echo /dev/mmcblk0p1 ;; esac; }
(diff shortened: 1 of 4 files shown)


igorpecovnik:
Hardware-confirmed on Banana Pi M7 (RK3588). Fresh eMMC install now boots from `/dev/mmcblk0p1`:

- partition table: **gpt** (was msdos)
- p1 start: **16.0MiB** (was 1MiB)
- u-boot intact on eMMC: `RKNS` @ sector 64, FIT (`d00dfeed`) @ sector 16384
- `rootdev` in armbianEnv.txt matches the eMMC p1 UUID
- `systemctl is-system-running` → running; Armbian 26.8.1, kernel 6.18.42-current-rockchip64

All three issues (boot0/boot1 menu #962, 16MiB offset, GPT table replication) are resolved end-to-end.

armbianworker (approved):

igorpecovnik:
Superseded by #972 (combined, unstacked so it can be merged — GitHub's stacked-PR feature blocked merging this via API).

PR #970, opened 2026-08-06
armbian-install: "boot from eMMC, root on NVMe/SATA/USB" (split install)

igorpecovnik:
> **Stacked on #963** (uses its `install_source_table_type`). Base is the #963 branch so the diff shows only this feature; retarget to `main` once #963 merges.

## What

Restores the classic installer's "Boot from eMMC – system on SATA/USB/NVMe". RK3588 and similar SoCs can't load u-boot from NVMe/SATA/USB, so a root on those needs the bootloader on internal storage. The new engine only offered `sd` mode (boot stays on the current removable media) for such targets — no way to pick an internal eMMC.

Adds a **`split-emmc`** scenario:
- **eMMC (boot device):** u-boot in the 16 MiB gap + a dedicated ext4 `/boot` the board's u-boot can always read + the remainder as a data partition auto-mounted at **`/emmc_storage`**.
- **target (NVMe/SATA/USB):** rootfs only. Its fstab mounts `/boot` from the eMMC boot partition; the eMMC boot env points `rootdev` at the target.

## Changes
- **engine:** new `emmc-boot` planner layout (`boot:512MiB:ext4` + `storage:100%:ext4`, 16 MiB start, source table type) and `install_run_split` composing the existing primitives across two devices.
- **frontend:** `split-emmc` offered **only** for nvme/sata/usb targets **when an eMMC is present** (`partitioner_emmc_device` detects it via boot0 / device type) and isn't the target; TUI warns *both* devices are erased; CLI `--boot split-emmc` (auto-detects the eMMC).
- **tests:** `emmc-boot` planner layout + table-type; loopback apply yields `/boot` + `/emmc_storage`.

## Validation
Local: `plan.bats` 25/25, `detect.bats` 7/7, `bootconfig.bats` 26/26; no shellcheck warnings on new code.

On a Banana Pi M7 (RK3588):
- menu offers `[sd split-emmc]` for the NVMe target (eMMC/SD targets still `[emmc sd]`);
- eMMC auto-detected (`/dev/mmcblk0`), source table `gpt`;
- real `parted` apply of the `emmc-boot` plan → p1 `16.0–528MiB` (512 MiB, legacy_boot) `/boot`, p2 `528MiB–end` `/emmc_storage`, GPT.

**Pending:** the full destructive `install_run_split` (rsync + u-boot write + actual boot from eMMC→NVMe) still needs a hardware run — requires wiping eMMC+NVMe and setting boot order / removing the SD to confirm.

## Follow-up
SPI/MTD-boot for a root-on-NVMe install: the engine's existing `mtd` mode already covers it; only the menu wiring (offer `mtd` as a boot choice for nvme/sata/usb targets, gated on an MTD device) is needed. Left out here to keep scope tight.

### DIFF
--- a/tests/bats/integration_loopback.bats
+++ b/tests/bats/integration_loopback.bats
@@ -52,6 +52,18 @@ teardown() {
 	LC_ALL=C parted -sm "$LOOP" unit MiB print | grep -qE '^1:16\.0MiB:'
 }
 
+@test "loopback: emmc-boot plan yields /boot + /emmc_storage, p1 at 16MiB" {
+	local plan; plan="$(install_plan_layout emmc-boot ext4 0 $((8*1024*1024*1024)) 512 0 gpt)"
+	run install_apply_partitions "$LOOP" "$plan"
+	[ "$status" -eq 0 ]
+	[[ "$output" == *"boot ${LOOP}p1"* ]]
+	[[ "$output" == *"storage ${LOOP}p2"* ]]
+	[ -b "${LOOP}p1" ]
+	[ -b "${LOOP}p2" ]
+	# boot partition clears the u-boot region and is ~512MiB; storage fills the rest
+	LC_ALL=C parted -sm "$LOOP" unit MiB print | grep -qE '^1:16\.0MiB:528'
+}
+
 @test "loopback: apply_partitions echoes role->device for each partition" {
 	local plan; plan="$(install_plan_layout uefi ext4 1 $((8*1024*1024*1024)) 512 0)"
 	run install_apply_partitions "$LOOP" "$plan"
--- a/tests/bats/plan.bats
+++ b/tests/bats/plan.bats
@@ -143,6 +143,28 @@ setup() {
 	[[ "$output" != *"part=swap:"* ]]
 }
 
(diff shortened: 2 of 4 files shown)


igorpecovnik:
Hardware-confirmed on Banana Pi M7 (RK3588), **btrfs root** (harder path — separate boot + initramfs rebuild). Fresh `--boot split-emmc --target /dev/nvme0n1 --fs btrfs` install, SD removed, booted standalone:

- **root** `/dev/nvme0n1p1` btrfs
- **/boot** `/dev/mmcblk0p1` ext4 (eMMC, 512 M)
- **/emmc_storage** `/dev/mmcblk0p2` ext4 (eMMC, 57.7 G)
- u-boot `RKNS` @ eMMC sector 64; `cmdline root=UUID=…` = NVMe, `rootfstype=btrfs`
- fstab: root=NVMe, /boot=eMMC, /emmc_storage=eMMC (nofail)
- `systemctl is-system-running` → running; **5.8 s** boot; no SD present

The full destructive path (partition both devices → rsync → u-boot write → boot from eMMC→NVMe with SD removed) is validated end-to-end.

armbianworker (approved):

igorpecovnik:
Superseded by #972 (combined with #963 into one unstacked PR).

PR #973, opened 2026-08-06
armbian-install: wire up "boot from SPI/MTD, root on NVMe/SATA/USB"

igorpecovnik:
## What

Makes the engine's existing `mtd` boot mode (u-boot on SPI/MTD flash, rootfs on the target) actually reachable — for boards like the **Odroid M1 (RK3568 SPI + NVMe)** that boot from SPI NOR.

## Why it didn't work before
- The boot-device menu only offered `mtd` when the **target itself** was an mtd device — never valid for a root target — so it never appeared for an NVMe/SATA/USB target.
- **`INSTALL_MTD_LIST`** (which the engine passes to `write_uboot_platform_mtd` to know where to flash u-boot) was **never populated** by the frontend, so even `--boot mtd` on the CLI would flash to an empty device.

## Changes
- **`partitioner_mtd_list()`** — detects the SPI/MTD device+partition list (`mtdblock*` whole devices + `/proc/mtd` `spl|boot|uboot` partitions as `mtdN:label`), mirroring the classic installer's `mtdcheck`.
- **Menu** — offer `mtd` for nvme/sata/usb targets when `write_uboot_platform_mtd` exists **and** an MTD device is present (removed the dead `role==mtd` case).
- **Dispatch** — export `INSTALL_MTD_LIST` for mtd installs in both TUI and CLI (CLI errors clearly if no MTD device found); TUI warns the SPI bootloader will be overwritten.
- **Engine** — `mtd` now also inherits the source partition-table type (its `/boot` lands on the target and is read by the SPI u-boot; Rockchip vendor u-boot is GPT-only).

## Validation
- Syntax + `plan.bats`/`detect.bats`/`bootconfig.bats` green; MTD detection produces the expected `mtdblock0 mtd0:uboot` → writes to `/dev/mtdblock0`; menu gate offers `mtd` only when hook + device present.
- **Hardware test pending on the Odroid M1** (below).

### DIFF
--- a/tests/bats/detect.bats
+++ b/tests/bats/detect.bats
@@ -95,6 +95,16 @@ setup() {
 	[ "$output" = "msdos" ]
 }
 
+@test "detect: excludes SPI/MTD flash (mtdblockN) - a boot device, not a target" {
+	# Odroid M1: 5 tiny mtdblock SPI partitions + one NVMe. Only the NVMe is a
+	# valid root destination; the SPI is offered later via the mtd boot mode.
+	run install_detect_targets "" "$(cat "$FIX/lsblk_mtd.json")"
+	[ "$status" -eq 0 ]
+	[ "${#lines[@]}" -eq 1 ]
+	[[ "$output" != *"mtdblock"* ]]
+	echo "$output" | grep -qP '^nvme0n1\tnvme\t'
+}
+
 @test "detect: surfaces sector size and model" {
 	run install_detect_targets "" "$(cat "$FIX/lsblk_4kn_large.json")"
 	# 4Kn NVMe reports phy-sec 4096 in field 4.
--- a/tests/bats/fixtures/lsblk_mtd.json
+++ b/tests/bats/fixtures/lsblk_mtd.json
@@ -1,0 +1,10 @@
+{
+    "blockdevices": [
+        {"name":"mtdblock0","type":"disk","size":917504,"phy-sec":512,"tran":null,"rota":false,"model":null},
+        {"name":"mtdblock1","type":"disk","size":131072,"phy-sec":512,"tran":null,"rota":false,"model":null},
+        {"name":"mtdblock2","type":"disk","size":2097152,"phy-sec":512,"tran":null,"rota":false,"model":null},
+        {"name":"mtdblock3","type":"disk","size":1048576,"phy-sec":512,"tran":null,"rota":false,"model":null},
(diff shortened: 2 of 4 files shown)


igorpecovnik:
Addressed the review findings in `0f6d14e8`. I placed the guards in the **shared engine entrypoints** rather than duplicating per-frontend, so `--target …`, the TUI selection, and the split path are all covered in one place:

**Fixed (valid):**
- **Reject `/dev/mtdblockN` as a root target** — `install_run_scenario` (and `install_run_split` for its root device) now refuse an MTD whole-device before `install_apply_partitions` wipes anything. Covers both the CLI `--target` and TUI inputs the reviewer flagged, without a CLI-only check.
- **Reject empty `INSTALL_MTD_LIST` before wiping** — `install_run_scenario` enforces it in pre-flight for `mtd` mode, which closes the TUI gap (the TUI recomputed the list but only the CLI checked it). The CLI's own check stays as a fast, clear error.
- **Docstring role list** — `install_detect_targets` no longer emits an `mtd` role (mtdblockN is filtered), so I removed it from the documented role set and noted MTD is offered via the boot mode instead.
- **`write_uboot_platform_mtd` contract** — documented next to `partitioner_mtd_list`: first list entry = boot device (`/dev/<first>` with any `:label` stripped), full `INSTALL_MTD_LIST` passed as the last arg, both `mtdblockN` and `mtdN:label` forms supported.

Tests green (plan/detect/bootconfig); the two guard conditions verified directly.

igorpecovnik:
@coderabbitai these three are the same findings from the previous round — all resolved in `0f6d14e8`, but enforced in the **shared engine entrypoints** rather than inline in the TUI/CLI, so a single guard covers both frontends (and the split path):

- **Reject `/dev/mtdblockN` as a root target** → `install_run_scenario` [module_install_engine.sh:972] and `install_run_split` [module_install_engine.sh:1145], before `install_apply_partitions` touches the device. This catches CLI `--target`, TUI selection, and split — not just the CLI path.
- **Reject empty `INSTALL_MTD_LIST` before wiping** → `install_run_scenario` pre-flight [module_install_engine.sh:982] for `mtd` mode. This is exactly the "enforce this invariant in the engine before partitioning" option you offered, and it closes the TUI gap centrally. The CLI also keeps its own fast check [module_partitioner.sh:331].

I intentionally didn't duplicate the `mtdblock`/empty-list checks inline in the CLI/TUI, since the engine is the single chokepoint every caller goes through — adding per-frontend copies would be redundant. Guard conditions verified; plan/detect/bootconfig tests green.

PR #974, opened 2026-08-06
armbian-install: tell the user why dual-boot can't proceed (BitLocker vs dirty NTFS)

igorpecovnik:
## Problem

A `uefi-dualboot` install against a Windows volume that can't be shrunk — **BitLocker-encrypted** or **unclean** (hibernated / Fast Startup / unclean shutdown) — failed at the `ntfsresize` step with the reason buried in `/var/log/armbian-install.log`. The TUI just showed "Installation FAILED, see log", so the user had no idea what to do.

## Fix

`install_dualboot_blocker()` runs **before anything is touched** and, if Windows can't be shrunk, echoes a plain user-facing reason **plus the exact fix** — and distinguishes the two cases (BitLocker is checked **first**, since an encrypted volume also fails the ntfsresize probe and must not be mislabeled "dirty"):

- **BitLocker** — detected via lsblk/blkid fstype **or** the `-FVE-FS-` boot-sector signature (a plain NTFS volume has `NTFS` there):
  > Turn off Device Encryption / BitLocker: `manage-bde -off C:`
- **Unclean NTFS** — `ntfsresize -f --info` reports inconsistent:
  > Disable Fast Startup: `powercfg /h off` · Check the disk: `chkdsk /f C:` · Shut down fully.

Surfaced in all three entry points, before the size prompt / any change:
- **TUI** → `dialog_msgbox` ("Cannot dual-boot yet")
- **CLI** → message to stderr
- **engine** (`install_run_dualboot`) → backstop for direct callers, logs the reason

## Validation

- syntax + `plan`/`detect`/`bootconfig`/`integration_loopback` tests green; no shellcheck warnings on the new code.
- **Real hardware** (Armbian live ISO on a UEFI box with a hibernated Windows NVMe): the helper prints the **dirty** message (correctly, not BitLocker — boot sector reads `NTFS`, not `FVE-FS`) with the actionable steps on screen, and returns non-zero before any destructive step. The install also stays fail-safe (ntfsresize runs before any partition change).

The differentiation is the key ask: BitLocker and dirty need different fixes, and the user now sees the right one without opening a log.

### DIFF
--- a/tools/modules/functions/module_install_engine.sh
+++ b/tools/modules/functions/module_install_engine.sh
@@ -809,6 +809,50 @@ install_disk_has_bitlocker() {
 	printf '%s' "$json" | jq -e 'any(.blockdevices[]?.children[]?; (.fstype // "") | test("bitlocker"; "i"))' >/dev/null 2>&1
 }
 
+install_dualboot_blocker() {
+	# install_dualboot_blocker <disk>
+	# If the disk's Windows volume cannot be shrunk for a dual-boot install, echo
+	# a plain, user-facing reason AND the exact fix, and return non-zero. Prints
+	# nothing and returns 0 when the disk is ready. The frontends show this text
+	# directly (dialog / message) so the user is told what to do without reading
+	# the install log.
+	local disk="$1" wi win
+	wi="$(install_detect_windows "$disk")" || {
+		echo "No Windows/UEFI install was found on $disk - there is nothing to dual-boot alongside. Use a normal (erase) install instead, or pick the disk that has Windows on it."
+		return 1
+	}
+	win="$(sed -n 's/^windows=//p' <<<"$wi")"
+	# BitLocker vs merely-dirty are different fixes, so distinguish them and check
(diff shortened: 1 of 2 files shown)


igorpecovnik:
Fixed in the latest commit. The finding was valid: `install_detect_windows` selects the Windows volume by `fstype == "ntfs"`, so a **BitLocker** volume (fstype `BitLocker`) made it fail, and the blocker fell through to the generic *"no Windows found"* message instead of the decrypt-first fix.

`install_dualboot_blocker` now:
1. selects the Windows system partition **independently of filesystem type** — the largest `Microsoft basic data` partition minus WinRE recovery;
2. checks **BitLocker on that partition only** (lsblk fstype **or** the `-FVE-FS-` boot-sector signature), not by scanning every partition;
3. routes encryption to `manage-bde -off C:` **before** requiring a shrinkable NTFS layout.

Verified on hardware: fs-independent selection picks `/dev/nvme0n1p3`; clean NTFS → ready (rc 0), dirty → Fast Startup/chkdsk message. Skipped nothing else in the review.

PR #977, opened 2026-08-10
firmware: never leave the board without a kernel on switch/repository change

igorpecovnik:
## Problem

Switching kernel branch or repository (e.g. **stable → rolling**) could strand the board with **no kernel at all**. Seen in the lab: after a rolling switch the SD card had its kernel packages missing and the board no longer booted.

The install path in `module_armbian_firmware` removed the old kernel **before** the new one was on disk:

```sh
for pkg in $packages; do pkg_remove 'linux-image*' ...; done   # purge first
pkg_install $packages                                          # then install
```

Three ways that bricks a board:

1. **Purge runs first** → a window with zero kernels installed; any interruption/failure in between leaves nothing bootable.
2. The **`linux-image*` wildcard** also matches the kernel about to be installed.
3. `pkg_install` returns an **unreliable exit code** in the non-`read` dialog-gauge path (the gauge pipeline masks apt's real status), so a failed install right after the purge looked like success.

## Fix

Reorder to be brick-safe (the download-first step is unchanged):

1. **Install the new kernel first**, as one `apt-get` transaction from the local cache, with a trustworthy exit code — apt replaces same-named packages in place so a bootable kernel is present at every step; on failure the current kernel is left untouched and we bail out.
2. **Prune only the *other* kernels** a branch/family switch leaves behind, by **exact name excluding what we just installed** — never a `linux-image*` wildcard.
3. **Assert a bootable kernel remains** (installed `linux-image` package + an image in `/boot`); fail loudly instead of letting the caller reboot into a dead system.

The `repository` command now **propagates an install failure** (source list switched but kernel could not be reinstalled from the new repo) instead of reporting a clean switch the caller would reboot into.

Verified: `bash -n` clean; stale-selection logic unit-tested (branch switch prunes only the old branch; same-branch version bump prunes nothing).

## Related

Companion safety net on the test side: armbian/autotests#92 stops the pipeline when a board never returns from a reboot/power-cycle.

### DIFF
--- a/tools/modules/system/module_armbian_firmware.sh
+++ b/tools/modules/system/module_armbian_firmware.sh
@@ -264,11 +264,25 @@ function module_armbian_firmware() {
 				return 1
 			fi
 
+			# Confirm the replacement image is actually installed BEFORE pruning
+			# the others. A broken/incomplete repo could install linux-dtb /
+			# linux-headers but omit the linux-image, leaving `keep` with no image
+			# package — the prune below would then remove every kernel. Bail here
+			# (current kernel untouched) rather than strand the board.
+			local target_image="linux-image-${branch}-${linuxfamily}"
+			if ! dpkg-query -W -f='${db:Status-Status}\n' "$target_image" 2>/dev/null | grep -qx 'installed'; then
+				rm -f /etc/apt/preferences.d/armbian-upgrade-policy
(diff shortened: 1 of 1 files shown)


igorpecovnik:
@coderabbitai review

igorpecovnik:
@coderabbitai full review

PR #978, opened 2026-08-13
armbian-install: add standalone "flash bootloader (u-boot) only" option

igorpecovnik:
## What

Adds back a **standalone bootloader-flash** path to `armbian-install`: write the running system's u-boot to **SPI/MTD** or an **SD/eMMC** device, with **no partitioning and no rootfs transfer**.

Reported by @rpardini in armbian/build#10176 — the new `armbian-install` could only write u-boot as part of a full install, but the old `nand-sata-install` let you "just flash u-boot." That matters for bringup: running from NVMe, or first boot via Maskrom/ramboot, where you only need u-boot on internal flash — not a reinstall.

## How

Reuses the **exact** `install_write_bootloader` path the installer already uses — so it's **no riskier than the install's existing bootloader step**, and adds no new flashing logic.

- **`partitioner_flash_uboot_tui()`** — pick SPI/MTD or an SD/eMMC device, confirm (clear "overwrites the bootloader, no data touched" warning), flash the running system's u-boot.
- **`partitioner_menu()`** — new top-level picker on the interactive path: **Install** vs **Flash bootloader only**. Only shown when the board can actually write u-boot (`write_uboot_platform` / `write_uboot_platform_mtd` present); otherwise goes straight to the install TUI. x86/GRUB systems are unaffected.
- **CLI:** `armbian-install bootloader --target <mtd|/dev/mmcblkN> --yes` (refuses without `--yes`).

Targets are restricted to **SD / eMMC / MTD** — the only media a SoC bootrom loads u-boot from (never NVMe/SATA/USB).

## Testing

- `bash -n` clean; shellcheck clean on the new code (only pre-existing SC2155 warnings remain).
- Unit-smoke-tested `partitioner_cli_flash` arg handling: missing target → 64, no MTD hook → 72, non-block target → 64, real block dev without `--yes` → 64 (refuse), with `--yes` → calls `install_write_bootloader mode=sd`, `--target mtd` → `mode=mtd dev=/dev/mtdblock0`.
- **Validated on real hardware** ✅ — flashed successfully on both target paths: an SD/eMMC board (`write_uboot_platform`) and a SPI-boot / NVMe-root board (`write_uboot_platform_mtd`).

### Follow-up fixes from review

- CLI `--target` with no value no longer loops forever (arg-count guard).
- CLI block target restricted to whole `mmcblk*` disks, so an arbitrary block device (NVMe/SATA/partition) can't get u-boot dd'd over its partition table — matches the interactive flow and the SD/eMMC/MTD-only scope.
- Interactive menu is target-aware: a box with no spare install target (SPI+NVMe) goes straight to the u-boot flash; the SPI/MTD label and confirmation dialog were tidied up.

Closes the gap noted in armbian/build#10176.

### DIFF
--- a/tools/modules/system/module_partitioner.sh
+++ b/tools/modules/system/module_partitioner.sh
@@ -387,23 +387,37 @@ partitioner_flash_uboot_tui() {
 }
 
 # Top-level action picker for the interactive installer: full install vs the
-# standalone bootloader flash. The bootloader-only entry only appears when the
-# board can actually write u-boot; otherwise go straight to the install TUI.
+# standalone bootloader flash. Built from what is ACTUALLY actionable on this
+# machine - so a board with no spare install target (e.g. booting from SPI with
+# root on NVMe, nothing left to install onto) just offers the u-boot update, and
+# a board that can only install (x86, no u-boot) goes straight to install. Only
+# shows the menu when both are possible; runs the single option directly otherwise.
 partitioner_menu() {
 	local title="Armbian installer"
-	if [[ "$(type -t write_uboot_platform)" == function || "$(type -t write_uboot_platform_mtd)" == function ]]; then
-		local act
-		act=$(dialog_menu " $title " "\nWhat would you like to do?" 0 78 4 -- \
-			install    "Install Armbian to a disk (SD / eMMC / NVMe / SATA / USB)" \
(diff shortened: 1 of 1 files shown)


rpardini:
Nice, thanks!

PR #981, opened 2026-08-15
install: keep the Rockchip reserved window

sormy:
`install_apply_partitions()` zeroes the first 10 MiB of the target. On Rockchip,
sectors 7168-16383 hold per-unit data that nothing recreates: vendor storage
(MACs, serial) tagged `DVKR` at 7168, secure storage (HDCP/DRM keys) tagged
`SSKR` at 8192. `rockchip64_common.inc` already names that map for SPI loader
images (`vnvm` 7168-7679, `reserved2` 8192-16383).

The loss is silent: `dwmac-rk` then finds no `LAN_MAC`, so U-Boot invents one with
`net_random_ethaddr()` and writes it back — the board keeps a stable address that
is not the one on its label.

**Implementation.** Read the two tags before wiping. If either is present, clear
the loader areas either side of the window (sectors 0-7167 and 16384-20479)
instead of the whole 10 MiB — same end sector, no temp files. If neither is
present the original `dd bs=1M count=10` runs unchanged, so this is a no-op on
any target with nothing there. A data check rather than a platform check: no
board list to maintain, and it fails safe toward current behaviour.

Three loopback tests, each checked to discriminate rather than pass trivially:

| scenario | asserts | fails when |
| --- | --- | --- |
| both tags | window survives, loaders cleared | unpatched |
| no tags | full 10 MiB wipe | — passes either way, pins unchanged behaviour |
| one tag only | window survives | gate mutated from OR to AND |

Suite is 11/11 with the fix, on an RK3528 board against a real loop device.

### DIFF
--- a/tests/bats/integration_loopback.bats
+++ b/tests/bats/integration_loopback.bats
@@ -39,6 +39,72 @@ teardown() {
 	[ -b "${LOOP}p2" ]
 }
 
+@test "loopback: emmc plan keeps the Rockchip reserved window, clears the loaders" {
+	# Sectors 7168-16383 hold vendor storage (MACs, serial) and secure storage
+	# (HDCP/DRM keys), provisioned per unit and never recreated. The loader areas
+	# either side of them must still be cleared. Markers sit on both edges so an
+	# off-by-one cannot pass.
+	printf 'IDB'   | dd of="$LOOP" bs=512 seek=64    conv=notrunc status=none
+	printf 'LDR0'  | dd of="$LOOP" bs=512 seek=7167  conv=notrunc status=none
+	printf 'DVKR'  | dd of="$LOOP" bs=512 seek=7168  conv=notrunc status=none
+	printf 'SSKR'  | dd of="$LOOP" bs=512 seek=8192  conv=notrunc status=none
+	printf 'RVEN'  | dd of="$LOOP" bs=512 seek=16383 conv=notrunc status=none
+	printf 'UBOOT' | dd of="$LOOP" bs=512 seek=16384 conv=notrunc status=none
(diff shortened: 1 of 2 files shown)


sormy:
Thanks — addressed in d1f84af.

**Preserving the range on non-Rockchip targets.** Fair; it now detects the store
tags first and only takes the narrower path when one is present (`DVKR` at 7168,
`SSKR` at 8192). Anything else gets the original 10 MiB `dd`, so non-Rockchip
installs are byte-for-byte unchanged. Added a test asserting exactly that.

**Boundary coverage.** Added, with your sector choices: 7167 and 20479 must be
cleared, 16383 must survive, 20480 must be untouched.

**Ignoring wipe failures.** The `|| true` on that `dd` is pre-existing and
unchanged here — happy to look at it separately, but tightening error handling on
the wipe seemed out of scope for this fix.

Verified on an RK3528 board with `bats` against a real loop device:

- patched: 10/10 pass
- unpatched: only `not ok 2 ... reserved window` fails; the new tagless-target
  test passes both ways, which is the point — it pins the unchanged behaviour.

sormy:
Good catch — the previous test wrote both tags, so an implementation requiring
both would have passed it. Added a case that drives each tag on its own.

Confirmed it actually discriminates: mutating the gate from OR to AND makes only
that case fail (`not ok 4 ... either Rockchip tag alone`), while the combined
case still passes — which is precisely the gap you described.

Suite is now 11/11 on an RK3528 board against a real loop device.

Note the branch has been squashed to one commit, so the `d1f84af` referenced in
my earlier reply is now folded into `01a914d`.

igorpecovnik (approved):

PR #985, opened 2026-08-17
install-engine: support extlinux boards when repointing root

igorpecovnik:
## Problem

Boards that boot via extlinux — spacemit/riscv and other vendor kernels — ship no `/boot/armbianEnv.txt`. Their kernel command line lives on the `append` line of `/boot/extlinux/extlinux.conf`. The engine only knew the armbianEnv `key=value` format, so on those boards:

- **`sd` mode** (rootfs to NVMe/USB, boot left on the removable media) aborted with `sd mode but current boot env (/boot/armbianEnv.txt) is missing; cannot make <disk> bootable` — a safe refusal, but it makes SD→NVMe installs impossible;
- **`emmc`/`mtd`/`ufs` and the eMMC split scenario** skipped the rewrite silently. Because `install_verify_boot_dir` already accepts `extlinux.conf` as a valid boot script, the install would report success while `root=` still pointed at the source media.

## Change

- `install_rewrite_extlinux()` — substitutes `root=` on the `append` line (adds it when absent), and `rootfstype=` the same way. Idempotent, preserves file mode, leaves `label`/`kernel`/`initrd`/`fdt` lines untouched.
- `install_boot_cfg_file()` / `install_rewrite_bootcfg()` — find and dispatch to whichever format the board actually uses. `armbianEnv.txt` keeps precedence when both exist, so no behaviour change on existing boards.
- Wired into all three call sites (`emmc|mtd|ufs`, `sd`, eMMC split).
- btrfs roots also get `rootflags=subvol=@`. armbianEnv boards get that from `boot.cmd`; extlinux boards never run it, so a btrfs install would otherwise mount the wrong subvolume.

## Testing

13 new bats cases in `tests/bats/bootconfig.bats` covering rewrite, idempotency, uppercase `APPEND`, missing `root=`, mode preservation, discovery precedence and dispatch. Fixture is a real `extlinux.conf` from a SpacemiT MUSE Book (K1, RISC-V).

Nothing has been flashed to hardware yet.

### DIFF
--- a/tests/bats/bootconfig.bats
+++ b/tests/bats/bootconfig.bats
@@ -73,6 +73,138 @@ setup() {
 	[ "$status" -eq 71 ]
 }
 
+# --- extlinux rewrite --------------------------------------------------------
+#
+# Distro-boot boards (spacemit/riscv and other vendor kernels) ship no
+# armbianEnv.txt; root= lives on the extlinux "append" line instead.
+
+# A real /boot/extlinux/extlinux.conf as written for a SpacemiT MUSE Book.
+_extlinux_fixture() {
+	cat >"$1" <<-'EOF'
+		label Armbian
+		  kernel /boot/Image
+		  initrd /boot/uInitrd
+		  fdt /boot/dtb/spacemit/k1-musebook.dtb
+		  append root=UUID=old-uuid console=ttyS0,115200 loglevel=1 rw splash
+	EOF
+}
+
+@test "extlinux: root= on the append line is repointed" {
+	f="$TMP/extlinux.conf"; _extlinux_fixture "$f"
+	run install_rewrite_extlinux "$f" "UUID=new-uuid" ext4
+	[ "$status" -eq 0 ]
+	grep -q 'root=UUID=new-uuid' "$f"
+	! grep -q 'old-uuid' "$f"
+	# every other kernel arg survives, in place
+	grep -q 'console=ttyS0,115200' "$f"
+	grep -q 'splash' "$f"
+}
+
+@test "extlinux: label/kernel/initrd/fdt lines are left alone" {
+	f="$TMP/extlinux.conf"; _extlinux_fixture "$f"
+	install_rewrite_extlinux "$f" "UUID=new-uuid" ext4
+	grep -q '^label Armbian$' "$f"
+	grep -q '^  kernel /boot/Image$' "$f"
+	grep -q '^  fdt /boot/dtb/spacemit/k1-musebook.dtb$' "$f"
+	[ "$(grep -c 'root=' "$f")" -eq 1 ]
+}
+
+@test "extlinux: rootfstype is added when absent and replaced when present" {
+	f="$TMP/extlinux.conf"; _extlinux_fixture "$f"
+	install_rewrite_extlinux "$f" "UUID=new-uuid" ext4
+	grep -q 'rootfstype=ext4' "$f"
+	# second pass with a different fs must not duplicate the token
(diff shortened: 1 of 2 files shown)


SuperKali:
LGTM. Tested on a Banana Pi R3 Mini (MT7986A, extlinux, ext4 root on eMMC): `root=` repointed, cmdline intact, idempotent.

igorpecovnik:
@SuperKali thanks for the R3 Mini test. The branch has moved since then, so here is what changed and what I think still needs a board before this merges.

### Rebased onto main (`bd97b39e`)

The PR was conflicting: main had gained the Raspberry Pi `sd`/`native` work that touches the same `sd` case this branch rewrites. Main's structure is kept (RPi detection, `native` mode) and only the non-RPi path was made extlinux-aware.

That required reworking the test seam. `install_sd_env_file` hardcoded `/boot/armbianEnv.txt`, so `install_sd_capable` would have hidden `sd` mode from extlinux boards entirely. It now resolves through `install_boot_cfg_file` over a new `install_sd_boot_dir` seam. **Consequence worth calling out: `sd` mode is now offered on extlinux boards where the menu previously hid it.** That is the intent, but it is a new path to real hardware.

### Review fixes (`583f995a`, `785e5bbb`)

- A `rootflags=subvol=@` inherited from a previous btrfs root survived a rewrite to ext4/f2fs. A non-btrfs kernel rejects `subvol=@` and the root mount fails. The bats case already walked into this (btrfs, then ext4) but only asserted the exit status, so it passed CI.
- `emmc|mtd|ufs` chained discovery and rewrite with `&&`, discarding the rewrite's exit status: a failed rewrite let the install continue to bootloader writing and report success with `root=` still pointing at the source. It now fails the scenario, as the `sd` and split paths already did. A *missing* boot config stays non-fatal.
- `install_rewrite_extlinux` returned success on a config with no `append` line at all — nothing rewritten, but the caller read it as a repointed target. It now returns `INSTALL_EX_BOOTCFG`, matching `install_rewrite_rpi_cmdline`, which already treats "no `root=` to rewrite" as an error.

### A regression that was introduced and fixed in the same branch (`d4dd9e3b`)

Worth flagging rather than burying. The first pass at the `rootflags` fix dropped the **entire** token on any non-btrfs rewrite. `rootflags` is not btrfs-only here — `aml-s9xx-box` ships `rootflags=data=writeback` in its stock `extlinux.conf` template, and star64 / visionfive2 / nezha carry the same in `SRC_CMDLINE`. Installing any of them to an ext4 root would have silently dropped a mount option the board expects, and an extlinux board with an ext4 root is exactly what this PR is for.

It now strips only the `subvol=` component and keeps the rest, removing the token entirely just when `subvol=` was all it held.

### Test status

117 bats tests pass. Each new test was checked to fail without its fix, so they are regression tests rather than decoration. But these are pure string transforms — the suite cannot tell you whether a board still boots.

What I do **not** think is covered by the R3 Mini run:

| Path | Why it needs a board |
| --- | --- |
| extlinux + board-shipped `rootflags` (aml-s9xx-box) | the regression above; ext4 root must keep `data=writeback` |
| btrfs → ext4 reinstall on extlinux | the original finding; stale `subvol=@` must be gone |
| `sd` mode on an extlinux board | newly reachable, never run on hardware |

The `emmc` + extlinux + ext4 path you tested is unchanged on success — the `&&` fix only alters behaviour when the rewrite *fails*.

If anyone has an aml-s9xx-box or a RISC-V board with `rootflags=data=writeback`, an install to ext4 plus a check that the flag survived in the installed `extlinux.conf` would cover the riskiest row.

PR #990, opened 2026-08-22
runners: authenticate the release lookup and verify before removing anything

igorpecovnik:
Explains and fixes:

```
jq: error (at <stdin>:1): Cannot index object with number
########################################################################## 100.0%
Removing runner insa-trixie-01 on GitHub
Removing runner 01 locally
gzip: stdin: not in gzip format
tar: Child returned status 1
-bash: line 1: ./config.sh: No such file or directory
```

## The chain

`module_armbian_runners.sh` looked up the runner version with an **unauthenticated** call:

```bash
LATEST=$(curl -sL https://api.github.com/repos/actions/runner/tags | jq -r '.[0].zipball_url' | ...)
```

api.github.com allows 60 requests an hour per IP unauthenticated. Past that, the reply is an error *object*, not the expected *array*:

| step | result |
|---|---|
| `jq -r '.[0].zipball_url'` on an object | `Cannot index object with number`, `LATEST` empty |
| URL becomes `…/download/v/actions-runner-linux-x64-.tar.gz` | 404 |
| `curl` without `--fail` | writes the 9-byte `Not Found` body, shows **100%** |
| `tar xzf` | `gzip: stdin: not in gzip format` |
| nothing extracted | `./config.sh: No such file or directory` |

Reproduced end to end locally; every line of the reported output is accounted for.

## Why it was worse than a failed install

The loop removes each runner **from GitHub and locally before extracting**. So a rate-limit blip doesn't just fail — it deletes the fleet and then cannot rebuild it. That is what the marching `Removing runner … 01 … 02 … 03` lines are.

## Change

1. **Authenticate** with the same token the registration-token calls already use, when one is supplied (`${gh_token:+-H "Authorization: Bearer …"}` — still works unauthenticated for a one-off run).
2. **`releases/latest` → `.tag_name`** instead of `tags[0]`, which relied on the order tags come back in.
3. **Verify before destroying**: the version must match `^[0-9]+\.[0-9]+\.[0-9]+$`, the download must succeed (`--fail`), and the archive must pass `tar tzf` — all before the first runner is touched. Each failure returns with *no runners were touched*.

## Verification

| `LATEST` value | outcome |
|---|---|
| the jq error text | refused, nothing touched |
| empty | refused, nothing touched |
| `Not Found` | refused, nothing touched |
| `2.330.0` | accepted |

Live check of the new endpoint returns `2.336.0`.

### DIFF
--- a/tools/modules/system/module_armbian_runners.sh
+++ b/tools/modules/system/module_armbian_runners.sh
@@ -107,10 +107,45 @@ function module_armbian_runners () {
 			local temp_dir=$(mktemp -d)
 			trap '{ rm -rf -- "$temp_dir"; }' EXIT
 			[[ "$ARCH" == "x86_64" ]] && local arch=x64 || local arch=arm64
-			local LATEST=$(curl -sL https://api.github.com/repos/actions/runner/tags | jq -r '.[0].zipball_url' | rev | cut -d"/" -f1 | rev | sed "s/v//g")
-			curl --progress-bar --create-dirs --output-dir ${temp_dir} -o \
-			actions-runner-linux-${ARCH}-${LATEST}.tar.gz -L \
-			https://github.com/actions/runner/releases/download/v${LATEST}/actions-runner-linux-${arch}-${LATEST}.tar.gz
+			# Ask GitHub for the latest runner release, authenticated with the
+			# same token the registration calls use: unauthenticated api.github.com
+			# allows 60 requests an hour per IP, and once that is spent the reply
+			# is an error object rather than the expected array, which used to end
+			# as `jq: Cannot index object with number` and an empty LATEST.
+			#
+			# releases/latest rather than tags[0]: it states which release is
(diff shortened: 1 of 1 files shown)


igorpecovnik:
Valid finding, fixed — and the reason it matters is worth stating: `trap … EXIT` set inside a function is **shell**-scoped, not function-scoped. It fires when armbian-config itself exits, not when the function returns. So the temp dir survived every early return, and on success the ~200 MB runner tarball stayed for the rest of the session.

```bash
runner_temp_cleanup() { rm -rf -- "${temp_dir}"; trap - EXIT; }
```

Called before each of the three `return 1` paths (version lookup, download, archive check) and at the end of the install arm. The original trap stays as the backstop for an unexpected exit.

Verified against a harness reproducing the same shape:

| path | temp dir after return |
|---|---|
| early return with cleanup | removed |
| success with cleanup | removed |
| early return without (old behaviour) | **still present** |

PR #1009, opened 2026-08-31
tests: cover ZFS install on trixie, noble and resolute

igorpecovnik:
`module_zfs` had no unit test. Adds `tests/zfs.conf` in the same shape as the other module tests (`unbound.conf` etc.): remove, install, status, remove.

```sh
ENABLED=true
RELEASE="trixie:noble:resolute"
TESTARCH="arm64,amd64"
TESTNAME="ZFS install"
```

`TESTARCH` follows the module's own declaration — `["module_zfs,arch"]="x86-64 arm64"` — so no armhf or riscv64.

## :warning: Heads-up before merging: this is likely to fail on noble and resolute

I ran the module in the actual test images (`ghcr.io/armbian/repository-update:<release>-amd64`) before opening this:

| Release | `module_zfs install` | Why |
|---|---|---|
| trixie | **rc=0** | no matching kernel headers available, so DKMS *skips* the build: `WARNING: No kernel headers were found, skipping module build` |
| noble | **rc=1** | headers *are* installable, so DKMS really builds — and the build fails |
| resolute | **rc=1** | same |

The noble failure in full:

```
Error! Bad return status for module build on kernel: 6.8.0-138-generic (x86_64)
dpkg: error processing package zfs-dkms (--configure):
 installed zfs-dkms package post-installation script subprocess returned error exit status 10
```

So on trixie the test passes for the wrong reason (nothing is compiled), and on noble/resolute it fails because zfs-dkms is built against the **container host's** kernel rather than an Armbian one.

That may be genuinely worth catching — or it may mean ZFS can't be meaningfully unit-tested in a container and the test should be trixie-only, or gated. Since CI will run it on this PR anyway, I've left all three releases in as requested; the result will show whether the CI runners' kernels behave the same as my local ones.

Related: `module_zfs kernel_max` returns `<not set>` inside the container, so the `ZFS001` menu condition (`linux-version compare "${KERNELID}" le "$(module_zfs kernel_max)"`) can't gate anything there either.

### DIFF
--- a/tests/zfs.conf
+++ b/tests/zfs.conf
@@ -1,0 +1,11 @@
+ENABLED=true
+RELEASE="trixie:noble:resolute"
+TESTARCH="arm64,amd64"
+TESTNAME="ZFS install"
+
+testcase() {(
+	./bin/armbian-config --api module_zfs remove
+	./bin/armbian-config --api module_zfs install
+	./bin/armbian-config --api module_zfs status
+	./bin/armbian-config --api module_zfs remove
+)}


igorpecovnik:
Valid, and it was masking a real failure. Fixed in `96fd520`.

The first and last `remove` are setup and cleanup, so their status is not a result; `install` and `status` are, and the trailing cleanup must not overwrite them:

```sh
rc=0
./bin/armbian-config --api module_zfs remove  || true
./bin/armbian-config --api module_zfs install || rc=$?
if [ "$rc" -eq 0 ]; then
    ./bin/armbian-config --api module_zfs status || rc=$?
fi
./bin/armbian-config --api module_zfs remove  || true
exit "$rc"
```

Verified with a stubbed `armbian-config`:

| scenario | old shape | now |
|---|---|---|
| all succeed | 0 | 0 |
| **install fails** | **0** :x: | **1** :white_check_mark: |
| status fails | 0 :x: | 1 :white_check_mark: |

## :warning: This will turn the job red, and correctly so

The previous green run was not evidence the module worked. From `ZFS install (noble/amd64)`:

```
Setting up zfs-dkms (2.4.4-1arter97~ubuntu24.04.1) ...
Error! Bad return status for module build on kernel: 6.8.0-137-generic (x86_64)
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

`zfs-dkms` fails to build against the runner's kernel on noble and resolute. On trixie it passes only because no matching kernel headers are installable there, so DKMS **skips** the build entirely — the package installs without ever compiling anything.

So once this merges, expect trixie green and noble/resolute red. That is the accurate picture. Options, depending on what this test is meant to prove:

- **packaging only** — keep all three releases and accept that DKMS is a no-op where headers are absent; noble/resolute then need the build failure addressed or the releases dropped
- **trixie only** — `RELEASE="trixie"`, acknowledging a container cannot build an out-of-tree module for a kernel it is not running
- **gate on headers** — skip the build assertion when `linux-headers-$(uname -r)` is not installable

Happy to take whichever you prefer. Worth noting the same masking affects the other ~50 tests using this shape — the HedgeDoc test in #741 went green through a port collision and a database that never started for exactly this reason.

PR #1015, opened 2026-09-07
tests: add Proxmox VE install test (trixie only)

igorpecovnik:
## TL;DR

Adds `tests/proxmox.conf`, following the same shape as `tests/zfs.conf` from #1009, but scoped to Debian trixie.

## Why trixie only

`module_proxmox` refuses to install anywhere else — Proxmox VE 9 packages are built for Debian 13:

```sh
if [[ "${DISTROID}" != "trixie" ]]; then
	dialog_msgbox "${title} not supported" ...
	return 1
fi
```

`TESTARCH="arm64,amd64"` for the same reason: the module's own `arch` metadata is `x86-64 arm64`, and upstream ships no other architectures.

## What it covers

remove (clean slate) → install → status → remove → status must now fail.

Unlike #1009 this wraps the body in `set -e` (the technique documented in `tests/README.md`), so every step has to pass. Without it only the final command decides the result, and a broken install would still report green. The trailing `! ... status` asserts removal actually took `pve-manager` away, rather than just exiting 0.

Matrix expands to `trixie/amd64` and `trixie/arm64`.

### DIFF
--- a/.github/workflows/maintenance-unit-tests.yml
+++ b/.github/workflows/maintenance-unit-tests.yml
@@ -56,6 +56,10 @@ jobs:
 
       - name: "Make JSON ${{ steps.changed-files.outputs.all_changed_files }}"
         id: json
+        env:
+          # Consumed as an env var rather than interpolated into the script
+          # body, so a file name can never be treated as shell.
+          CHANGED_TESTS: "${{ steps.changed-files.outputs.all_changed_files }}"
         run: |
 
           # Run each enabled test case against every arch its TESTARCH
@@ -105,8 +109,15 @@ jobs:
           if [[ "$run_all_tests" == "true" ]]; then
               tests=($(grep -rwl tests/*.conf -e "ENABLED=true" | cut -d":" -f1))
           else
-              # Get changed test files using git to avoid parsing issues
-              changed_tests=($(cd "${GITHUB_WORKSPACE}" && git diff --name-only origin/main...HEAD | grep '^tests/.*\.conf$' || true))
+              # Take the list from tj-actions/changed-files above: it already
+              # resolved the PR base and fetched whatever history that needed.
+              # Re-deriving it here with `git diff origin/main...HEAD` dies with
+              # "no merge base" on the shallow default checkout, and the `|| true`
+              # turned that into an empty matrix that reported success.
+              changed_tests=()
+              while IFS= read -r changed_test; do
+                  [[ "${changed_test}" == tests/*.conf ]] && changed_tests+=("${changed_test}")
+              done <<< "${CHANGED_TESTS}"
               if [[ ${#changed_tests[@]} -gt 0 ]]; then
                   # Filter for ENABLED=true
                   tests=($(grep -rwl -e "ENABLED=true" -- "${changed_tests[@]}" 2>/dev/null | cut -d":" -f1))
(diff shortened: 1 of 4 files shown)


igorpecovnik:
First run failed on both arches — and it turned out the test was right.

The install itself worked; the log ends with the success message and then dies 40 ms later:

```
18:52:10.417  Proxmox VE is installed on the Armbian kernel.\n\nWeb UI: https://172.18.0.2:8006
18:52:10.456  ##[error]Process completed with exit code 1.
```

`dialog_msgbox` in `read` mode ended with `read -p "Press Enter to continue..."`, and that status became the function's. `--api` faithfully propagates a module's exit status, so under `--api` (stdin at EOF) `read` returns 1 and the whole install reports failure after doing everything correctly. `module_proxmox install` ends with exactly such a message box.

Reproduced locally:

```
$ dialog_msgbox "T" "installed ok" 10 66 </dev/null
installed ok
exit=1
```

A message box has nothing to cancel, so it should never fail. 0fd9aad skips the prompt when stdin is not a tty; interactive use is unchanged (verified under a pty — prompt still shown, exit 0).

This affects every module whose last statement is a message box, not just Proxmox.

igorpecovnik:
Second round of findings — the test is doing its job.

**Why the matrix went empty (7dc1aff)**

The run before last passed without executing anything. The matrix step derived its own list of changed tests with `git diff --name-only origin/main...HEAD`, but `Checkout` is shallow, so:

```
fatal: origin/main...HEAD: no merge base
```

`|| true` swallowed it, `tests=()` came out empty, and the workflow went green. `tj-actions/changed-files` runs immediately before and already had the right answer — the step is literally titled "Make JSON tests/proxmox.conf". Now it uses that output instead of recomputing it.

**The Proxmox kernel was being installed anyway (1339de77)**

The point of installing `pve-manager` rather than `proxmox-ve` was to keep the Proxmox kernel off an Armbian system. It didn't work:

```
pve-manager  --Depends-->  pve-yew-mobile-gui  --Recommends-->  proxmox-ve  --Depends-->  proxmox-default-kernel
```

apt installs Recommends by default, so `proxmox-kernel-7.0`, `proxmox-kernel-7.0.14-15-pve-signed` and `proxmox-default-kernel` all came along. That single Recommends is the only link, so an apt pin on `proxmox-ve` (plus the kernel packages, for future-proofing) cuts it.

Verified in a `debian:trixie` container against the real repo:

| | kernel/meta packages selected | pve-manager |
|---|---|---|
| without pin | 5 | installed |
| with pin | **0** | still installed |

**The enterprise repo**

Not something we configured — `pve-manager` itself ships `/etc/apt/sources.list.d/pve-enterprise.sources`, with no `Enabled:` key, so it is on by default:

```
Types: deb
URIs: https://enterprise.proxmox.com/debian/pve
Suites: trixie
Components: pve-enterprise
```

Without a subscription that answers `401 Unauthorized` and every later `apt update` fails. Now disabled after install (it is a conffile, so edited rather than deleted); the no-subscription repo we write is the one that stays.

**Which is what broke `remove`**

```
W: (pve-apt-hook) You are attempting to remove the meta-package 'proxmox-ve'!
E: Sub-process /usr/share/proxmox-ve/pve-apt-hook returned an error code (1)
```

`proxmox-ve` installs an apt hook that aborts any transaction removing it until `/please-remove-proxmox-ve` exists. With the pin, `proxmox-ve` never gets installed in the first place, but `remove` now creates the marker anyway so it still works on systems set up before this change.

